PR #190, opened 2022-04-08
Feature/ipfs

Peyton232:


highway/crud_object.go
```diff
@@ -8,6 +8,43 @@ import (
 
 // CreateObject creates a new object.
 func (s *HighwayServer) CreateObject(ctx context.Context, req *ot.MsgCreateObject) (*ot.MsgCreateObjectResponse, error) {
+	// res, err := json.Marshal(req)
```
mcjcloud: Why all the commented out code?

highway/server.go
```diff
@@ -96,28 +100,55 @@ func NewHighway(ctx context.Context, opts ...hn.Option) (*HighwayServer, error)
 	// purges expired items every 10 minutes
 	c := cache.New(5*time.Minute, 10*time.Minute)
 
+	// TODO work with Nick on what exact approach to do on this
```
mcjcloud, line 107: Will the code still work without this?

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/ipfs/go-ipfs/core"
+	"github.com/ipfs/go-ipfs/core/coreapi"
+	"github.com/ipfs/go-ipfs/core/node/libp2p"
+	"github.com/ipfs/go-ipfs/plugin/loader" // This package is needed so that all the preloaded plugins are loaded automatically
+	"github.com/ipfs/go-ipfs/repo/fsrepo"
+	"github.com/libp2p/go-libp2p-core/peer"
+	status "github.com/sonr-io/core/errors"
+)
+
+var FlagExp = flag.Bool("experimental", false, "enable experimental features")
```
mcjcloud, line 30: nit: A name like `enable_ipfs_experimental` would be less ambiguous in case there are also other experimental features down the road.

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	repoPath, err := createTempRepo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp repo: %s", err)
+	}
+
+	// Spawning an ephemeral IPFS node
+	return createNode(ctx, repoPath)
+}
+
+func ConnectToPeers(ctx context.Context, ipfs iface.CoreAPI, peers []string) error {
```
mcjcloud: nit: Add docstring comment

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	fmt.Printf("Added file to IPFS with CID %s\n", cidFile.String())
+
+	newLocation := ipfsStorageDir
+	os.Mkdir(newLocation, 0755)
+	if err != nil {
+		return UploadResponse{Status: status.StatusInternalServerError}, err
+	}
+	newLocation += cidFile.String()
```
mcjcloud: Use filepath.Join() here as it's safer and accounts for differences in filesystems and slashes.

https://pkg.go.dev/path/filepath#Join

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	os.Mkdir(newLocation, 0755)
+	if err != nil {
+		return UploadResponse{Status: status.StatusInternalServerError}, err
+	}
+	newLocation += "/bytes.txt"
```
mcjcloud: Same as above

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	Path      string
+}
+
+func DownloadData(ctx context.Context, cid string, node iface.CoreAPI) (DownloadResponse, error) {
+	outputBasePath, err := ioutil.TempDir("", "example")
+	if err != nil {
+		panic(fmt.Errorf("could not create output dir (%v)", err))
+	}
+
+	//TODO strip cid down to just the cid string portion
```
mcjcloud, line 274: What does this mean?

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+
+	path := outputBasePath + cid
```
mcjcloud: Same as above (use filepath.Join())

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+	cidPath := ifacepath.New(cid)
+
+	//check local file system first
+	if _, err := os.Stat(ipfsStorageDir + "/" + cid + "/bytes.txt"); !errors.Is(err, os.ErrNotExist) {
```
mcjcloud: You can also use filepath.Join() here

`filepath.Join(ipfsStorageDir, cid, "bytes.txt")`

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+	/// --- Part II: Upload a file to IPFS
+	data := []byte("Hello World!!!")
+	respUp, err := UploadData(ctx, data, node)
+	if err != nil {
+		t.Errorf("UploadData([]byte, coreAPI) resulted in status %d", respUp.Status)
+		t.Error(err)
+	}
+
+	/// --- Part III: Retrieve a file off of IPFS
+	bootstrapNodes := []string{
```
mcjcloud: Making these a top level constant would reduce code duplication and clean this function up

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// "/ip4/127.0.0.1/udp/4010/quic/p2p/QmZp2fhDLxjYue2RiUvLwT9MWdnbDxam32qYFnGmxZDh5L",
+	}
+	go func() {
+		err := ConnectToPeers(ctx, node, bootstrapNodes)
+		if err != nil {
+			log.Printf("failed connect to peers: %s", err)
+		}
+	}()
+
+	exampleCIDStr := "QmUaoioqU7bxezBQZkUcgcSyokatMY71sxsALxQmRRrHrj"
```
mcjcloud: Shouldn't the CID string come from `respUp`? Seems like the idea should be to test that you can download the same data that you uploaded.

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// You can add more nodes here, for example, another IPFS node you might have running locally, mine was:
+		// "/ip4/127.0.0.1/tcp/4010/p2p/QmZp2fhDLxjYue2RiUvLwT9MWdnbDxam32qYFnGmxZDh5L",
+		// "/ip4/127.0.0.1/udp/4010/quic/p2p/QmZp2fhDLxjYue2RiUvLwT9MWdnbDxam32qYFnGmxZDh5L",
+	}
+	go func() {
+		err := ConnectToPeers(ctx, node, bootstrapNodes)
+		if err != nil {
+			log.Printf("failed connect to peers: %s", err)
+		}
+	}()
```
mcjcloud: How does this relate to the rest of the test? It's likely this will not be done before the download is attempted. Is this also needed for the upload?

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+	/// --- Part II: Upload a file to IPFS
+	data := []byte("Hello World!!!")
+	respUp, err := UploadData(ctx, data, node)
+	if err != nil {
+		t.Errorf("UploadData([]byte, coreAPI) resulted in status %d", respUp.Status)
+		t.Error(err)
+	}
+
+	/// --- Part III: Retrieve a file off of IPFS
+	bootstrapNodes := []string{
```
mcjcloud: Same as above, use the constant

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// "/ip4/127.0.0.1/udp/4010/quic/p2p/QmZp2fhDLxjYue2RiUvLwT9MWdnbDxam32qYFnGmxZDh5L",
+	}
+	go func() {
+		err := ConnectToPeers(ctx, node, bootstrapNodes)
+		if err != nil {
+			log.Printf("failed connect to peers: %s", err)
+		}
+	}()
+
+	exampleCIDStr := "QmUaoioqU7bxezBQZkUcgcSyokatMY71sxsALxQmRRrHrj"
```
mcjcloud: Same as comments above

ipfs/ipfs_test.go
```diff
@@ -0,0 +1,133 @@
...
+	respDown, err := DownloadData(ctx, exampleCIDStr, node)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if respDown.Status != status.StatusOK {
+		t.Errorf("DownloadData(ctx, cid, node) resulted in not OK status")
+	}
```
mcjcloud, line 72: Consider also checking that the downloaded data matches the uploaded data "Hello World!!!" (Same thing below)

mcjcloud (changes requested):

errors/error.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2009 The Go Authors. All rights reserved.
```
joshLong145: Can we move this to a constants folder and name it statuses or statusCodes? there are more than just error codes in the file.

highway/server.go
joshLong145: I think it will crash without this configuration directory, we should probably uncomment this for now

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+
+/// ------ Spawning the node
+
+// Creates an IPFS node and returns its coreAPI
+func createNode(ctx context.Context, repoPath string) (iface.CoreAPI, error) {
+	// Open the repo
+	repo, err := fsrepo.Open(repoPath)
+	if err != nil {
+		return nil, err
+	}
```
joshLong145, line 104: might want to add some debug logging with the open operation fails.

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+		Repo:      repo,
+	}
+
+	node, err := core.NewNode(ctx, nodeOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	// Attach the Core API to the constructed node
+	return coreapi.NewCoreAPI(node)
```
joshLong145, line 121: instead of returning the api instance, can we use a variable and check for errors before return? that way we can shutdown on error of the api stand up

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+		return nil, err
+	}
+
+	f, err := files.NewSerialFile(path, false, st)
+	if err != nil {
+		return nil, err
+	}
+
+	return f, nil
+}
```
joshLong145, line 225: With this implementation i believe we are binding to a unix fs implementation. is there another option that ipfs supports? we might want to consider a "universal file system implementation" albeit probably not supported.

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+
+type UploadResponse struct {
+	Status int
+	Cid    string
+	Path   string
+}
```
joshLong145: can this type def be moved to its own file

ipfs/ipfs.go
joshLong145: agree with @mcjcloud

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+		Cid:    cidFile.String(),
+	}, nil
+}
+
+type DownloadResponse struct {
+	Status    int
+	FileData  []byte
+	NameSpace string
+	Path      string
+}
```
joshLong145: Same as above, move type def to its own file if possible.

ipfs/ipfs.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		return DownloadResponse{
+			Status:    status.StatusOK,
+			FileData:  data,
+			Path:      cidPath.String(),
+			NameSpace: cidPath.Namespace(),
+		}, nil
+	}
+
+	rootNodeFile, err := node.Unixfs().Get(ctx, cidPath)
```
joshLong145, line 293: are we checking the cidPath for validity before operating on the FS?

ipfs/ipfs_test.go
joshLong145: I agree, that would be a bit cleaner

joshLong145 (changes requested):

highway/server.go
Peyton232: It will work without this for ephemeral nodes which is what we are running currently. I will work with Nick at some point to work out the kinks with default nodes. I think this process here would be better done in a dockerfile but I'm leaving this here till that is taken care of.

ipfs/ipfs.go
Peyton232: This name seems a little wordier than I would like, but I do like it over FlagExp, which is a name I just stole off of their example. Either way I don't see us wanting to use experimental features for the time being. In my mind I imagined this would be for dev use only, if I'm wrong in that assumption than I think making this an env variable would be better to make it configurable. 

This flag just says if you want to use any experminetal features and if enabled just turns all of them on. We could make this process more sophisicated if we think it would be useful.

ipfs/ipfs.go
Peyton232: looking into fsrepo.open it seems like they have lots of internal checks that woudl get returned in err. It woudl be idiomatic to just log this err variable by the caller. I don't think extra debug logging would be particularly helpful if already covered in the .Open function (such as path not existing, permission mismatch, etc)

ipfs/ipfs.go
Peyton232: Currently the flow is returning the API and err to the spawner functions who then return that err to the caller, this follows the idiomatic error handlng in Go. 
We can not use this structure of error handling but I would like to have a brief sync to go over this change.

ipfs/ipfs.go
Peyton232: I personally did not what to bind to file ssystem at all but this seems to be the only way supported, I think it would be a good idea to expand the go-ipfs-as-a-library with data tsream support and/or universal file system implementation.

ipfs/ipfs.go
Peyton232: I just want the iodentifier part of the cid
example cid "/ipfs/iuwebfiubverwqqivbewviugev"
I only want that last section of the cid

ipfs/ipfs.go
Peyton232: There is some validity checks first thing in UnicFs().Get() if they fail then the reason why will be returned in err.

ipfs/ipfs_test.go
Peyton232: I can add that as a seperate test, this is a CID it gets from the actual network. If I tested for the uplaoded file from before it would grab it from the local filesystem since that would be the closest

ipfs/ipfs_test.go
Peyton232: The CID below is pulled from these peers.

Peyton232: If we wanted to uplload to this network then we would have to bootstrap before the upload I believe

ipfs/ipfs_test.go
Peyton232: we are pulling a different piece of data, we are pulling the exampleCID from the network we connected to on the bootstrappers

ipfs/ipfs.go
mcjcloud: Okay lgtm

ipfs/ipfs.go
mcjcloud: Agreed, the error will be returned from this method. Resolving

ipfs/ipfs_test.go
mcjcloud: Okay makes sense. My only question now is: Is it possible that it will try to fetch the CID before this completes? That would make the test fail right?

ipfs/ipfs.go
mcjcloud: Resolving.

ipfs/ipfs.go
mcjcloud: I like the idiomatic way of retuning errors from the function.

As an aside, do you plan to convert these functions into methods on a struct in the PR?

ipfs/ipfs.go
mcjcloud: Seems like for many tasks the file system is only an intermediary between data in memory and data in IPFS. We should look into a way to go straight from memory to IPFS and vice versa.

ipfs/ipfs.go
mcjcloud: Resolving.

ipfs/ipfs.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil, fmt.Errorf("failed to create temp repo: %s", err)
+	}
+
+	// Spawning an ephemeral IPFS node
+	node, err := createNode(ctx, repoPath)
+	return &IpfsNode{CoreAPI: node}, err
+}
+
+// Given a list of bootstrapper nodes try and connect to teh ipfs node
+func ConnectToPeers(ctx context.Context, ipfs iface.CoreAPI, peers []string) error {
```
mcjcloud: Method signature here can be updated to `func (n *IpfsNode) ConnectToPeers()`

ipfs/ipfs.go
```diff
@@ -0,0 +1,311 @@
...
+	f, err := files.NewSerialFile(path, false, st)
+	if err != nil {
+		return nil, err
+	}
+
+	return f, nil
+}
+
+// External Functions
+func (node IpfsNode) UploadData(ctx context.Context, data []byte) (UploadResponse, error) {
```
mcjcloud: This needs to be a pointer method (e.g. `func (node *IpfsNode)` because you're performing operations on `node` below.

mcjcloud (changes requested):

ipfs/ipfs.go
Peyton232: Yes, that's like I was saying I do not personally like this approach, but go-ipfs-as-a-library is built with the idea of going straight from file system to ipfs.

ipfs/ipfs_test.go
prdlk: This is a redundant bootstrapping process. Our libp2p node that is embedded in the HighwayServer service already handles it. If there is mechanism for utilizing our libp2p host interface opposed to running duplicate nodes on the server with same functionality - would be great

Peyton232: The get functionality will comtinue to search remote systems until it finds what it wants, this is why it is recommended to put a deadline in the ctx of this call when we use it.

joshLong145 (approved):
Looking good. approving

joshLong145:
Branch needs to be updated soon, per node standup of `sonard` binary

PR #197, opened 2022-04-18
Feature/object crud

Peyton232:
Adds basic Creation, Update, and deactivation functionality for object type


highway/crud_object.go
```diff
@@ -2,22 +2,158 @@ package highway
 
 import (
 	context "context"
+	"errors"
 
+	otv1 "github.com/sonr-io/blockchain/x/object/types"
```
joshLong145: rename to 'ot_v1'

highway/crud_object.go
```diff
@@ -2,22 +2,158 @@ package highway
...
+	}
+
+	// Broadcast the message
+	res, err := s.cosmos.BroadcastUpdateObject(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	// decode otv1 fields to ot
+	outputFields := s.decodeObjectDocFields(res.WhatIs.ObjectDoc.Fields)
```
joshLong145: should we first check for non nill properties on result to prevent panics

highway/crud_object.go
```diff
@@ -2,22 +2,158 @@ package highway
...
+}
+
+// -----------------
+// Helper Functions
+// -----------------
+
+// Translate ot (buf) fields to otv1 (blockchain) //TODO rename this
+func (s *HighwayServer) bufToBlockchain(bufFields []*ot.ObjectField) []*otv1.ObjectField {
+	var blockchainFields []*otv1.ObjectField
+	for _, v := range bufFields {
```
joshLong145, line 142: we should add some checks on the collection for before iterating over it.

highway/crud_object.go
```diff
@@ -2,22 +2,158 @@ package highway
...
+		blockchainFields = append(blockchainFields, item)
+	}
+
+	return blockchainFields
+}
+
+// Translate the objectDoc otv1 (blockchain) fields to ot (buf) //TODO rename this
+func (s *HighwayServer) decodeObjectDocFields(blockchainFields map[string]*otv1.ObjectField) map[string]*ot.ObjectField {
+	// decode otv1 fields to ot
+	outputFields := make(map[string]*ot.ObjectField, len(blockchainFields))
```
joshLong145, line 163: check first for the length before making the new object

joshLong145 (changes requested):

highway/crud_object.go
Peyton232: Should I go ahead and change the name for registery, channels, and buckets to match this scheme?

joshLong145: if its low impact we can migrate naming in this PR

highway/crud_object.go
Peyton232: Yes, cause len does not short circuit on nil pretty sure

highway/crud_object.go
Peyton232: What checks are you thinking?

joshLong145: nill pointer check and length check

Peyton232: I believe these would both be redundant with the range function, it will simply not iterate on nil or 0 len.

joshLong145 (dismissed):

highway/crud_object.go
```diff
@@ -2,22 +2,171 @@ package highway
...
+		return nil
+	}
+
+	outputFields := make(map[string]*ot.ObjectField, len(blockchainFields))
+	for k, v := range blockchainFields {
+		outputFields[k] = &ot.ObjectField{
+			Label: v.GetLabel(),
+			Type:  ot.ObjectFieldType(v.GetType()),
+			Did:   v.GetDid(),
+			//Value:    v.Value,
```
mcjcloud: nit: remove commented code

highway/crud_object.go
```diff
@@ -2,22 +2,171 @@ package highway
...
+
+	return blockchainFields
+}
+
+// Translate the objectDoc ot_v1 (blockchain) fields to ot (buf) //TODO rename this
+func (s *HighwayServer) decodeObjectDocFields(blockchainFields map[string]*ot_v1.ObjectField) map[string]*ot.ObjectField {
+	// decode ot_v1 fields to ot
+
+	if blockchainFields == nil || len(blockchainFields) < 1 {
+		return nil
```
mcjcloud: Returning an empty map would be better than `nil` here. There's not typically a need for nil maps.

mcjcloud (dismissed):

highway/crud_object.go
joshLong145: I would agree, keeps code paths nicer

highway/crud_object.go
```diff
@@ -2,22 +2,221 @@ package highway
...
+			Type:        types.WhoIs_Type(regSess.Whois.GetType()),
+			Metadata:    regSess.Whois.GetMetadata(),
+			Timestamp:   regSess.Whois.GetTimestamp(),
+			IsActive:    regSess.Whois.GetIsActive(),
+		},
+		Credential: &types.Credential{
+			ID:              regSess.Credential.GetID(),
+			PublicKey:       regSess.Credential.GetPublicKey(),
+			AttestationType: regSess.Credential.GetAttestationType(),
+			// Authenticator: &types.Authenticator{  //TODO this causes nil dereference, figure out why
```
joshLong145, line 196: Authenticator needs to be hydrated, this object is critical

highway/crud_object.go
joshLong145: we should guard so we can provide contextual information, perhaps it should return ([]*ot_v1.ObjectField, error)

joshLong145 (approved):
approving, noting that we should probably unify method signatures to return responses and errors in the future for better context

PR #201, opened 2022-04-19
feat: update login endpoints for did query

joshLong145:
## Describe your changes
- changes to crud_registry to now query on all registered whoIs objects for the given username to map to a did

### DIFF
--- a/highway/client/cosmos.go
+++ b/highway/client/cosmos.go
@@ -131,7 +131,7 @@ func (cc *Cosmos) QueryName(name string) (*rt.WhoIs, error) {
 	}
 
 	queryResp, err := cc.registryQuery.WhoIs(context.Background(), &rt.QueryWhoIsRequest{
-		Did: name,
+		Did: whoIsMatch.Did,
 	})
 	if err != nil {
 		golog.Errorf("Error querying name: %s", err.Error())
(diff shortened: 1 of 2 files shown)


highway/crud_registry.go
```diff
@@ -136,13 +136,27 @@ func (s *HighwayServer) StartAccessName(w http.ResponseWriter, r *http.Request)
...
+		if whoIs.Name == username {
+			who = &whoIs
+		}
+	}
+
+	if who == nil {
+		log.Println(err)
+		JsonResponse(w, errors.New("Error while querying did for domain name"), http.StatusBadRequest)
+		return
+	}
```
mcjcloud: I updated cosmos.go to perform this logic for querying name, take a look and see if you can just use `s.cosmos.QueryName(username)`

joshLong145: Just looked into queryName, and yes i agree. migated from an exhaustive query to the wrapper. Thanks

highway/client/cosmos.go
```diff
@@ -131,7 +131,7 @@ func (cc *Cosmos) QueryName(name string) (*rt.WhoIs, error) {
 	}
 
 	queryResp, err := cc.registryQuery.WhoIs(context.Background(), &rt.QueryWhoIsRequest{
-		Did: name,
+		Did: whoIsMatch.Did,
```
mcjcloud, line 134: Good catch 👌🏻

mcjcloud (approved):

PR #213, opened 2022-04-23
Objects should now support values and querying

mcjcloud:


joshLong145 (approved):
thank you, this looks good enough to get us moving forward. will need to address the on chain storage but that is an easy refactoring

PR #246, opened 2022-04-26
Migrate repos into a monorepo

mcjcloud:
Combine blockchain, core, and highway-sdk into a single repo, using https://github.com/golang-standards/project-layout as a reference.

## Changes

- See https://www.notion.so/sonr/Monorepo-Migration-509bd03aee6f45d1b397554fa14fd2a2 for file structure


ntindle:
Where is the highway being served from?

ntindle:
Also may be missing it but I don't see most of the chain related stuff anymore

joshLong145:
I dont think we want to use sub modules. I think using go module format is the way we want to move forward, which keeps us a bit more generic.

mcjcloud:
> I dont think we want to use sub modules. I think using go module format is the way we want to move forward, which keeps us a bit more generic.

This was done by mistake, will figure out how to revert

ntindle:
May want to rewrite readme

prdlk:
Oh baby this is it isnt it

ntindle:
This looks like it stands up but I can't get the endpoints to work. 

~Getting 404 not found on all endpoints~
NVM I can't type

ntindle:
Looks like blob remove doesn't work for real though

ntindle (dismissed):

ntindle (approved):

ntindle:
LGTM, swagger taskfile needs updated to output to the correct directory

PR #254, opened 2022-04-29
Clean/docs structure

prdlk:
Description of PR...

## Changes

* Shifted Blockchain docs to root folder
* Moved vue wallet into web folder
* Changed config.yaml to resolve updated paths
* Updated Taskfile to follow colon notation
* Added Goreleaser CI/CD to build cli and provide releases
* Added Taskfile to web projects
* Moved Host into internal, our packaged highway and motor leverage host and does not need to be exposed

### `pkg/channel`

Moved all NewChannel logic into `internal/blockchain/x/channel/types` for standard implementation and removing duplication


Taskfile.yaml
```diff
@@ -1,36 +1,76 @@
...
+  build:swagger:
+    cmds:
+      - echo "Building swagger api docs for highway"
+      - swag i -g pkg/highway/server.go --output docs/highway
+      - echo "Building Open api docs for Blockchain"
+      - starport generate openapi
+      - starport generate proto-go
+      - go mod tidy
+
+  chain:serve:
```
mcjcloud: nit: You have `build:chain` but `chain:serve`. I think these should be made consistent

web/wallet/Taskfile.yaml
```diff
@@ -19,4 +21,5 @@ tasks:
 
   serve:
     cmds:
-      - npm run dev
+      - npm install
```
ntindle: This command has side-effects. I would not recommend doing this as part of other steps without being clear

Taskfile.yaml
prdlk: okay serve:chain

web/wallet/Taskfile.yaml
prdlk: Okay removed im not a web dev XD

mcjcloud (approved):
lgtm

Issue #285, opened 2022-05-06
[BUG] Webauthn Session Store not persistent


mcjcloud:
**Describe the bug**
Somewhere in between the start registration and finish registration the webauthn session store is deleted.

**To Reproduce**

1. Run highway
2. Breakpoint after webauthn session store is set on start request
3. Breakpoint before webauthn session store is read on finish request
4. Run registration and observe the value is set after the first breakpoint but not at the second.

**Expected behavior**
The value should exist during finish registration.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202534238572263) by [Unito](https://www.unito.io)

PR #289, opened 2022-05-06
Draft: Fix sessions

mcjcloud:
Fixes the auth workflow with sessions and enable APIs to work.

## Changes

- Register and login CLI commands
- Opened #285 to address issues with Webauthn session storage
- Make use of session in highway API calls

### New Features

Calls to highway should all complete successfully without Sessions being a blocker.

## Checklist

- [x] CLI register
- [x] Browser-based auth
- [x] Unit tests
- [x] Documentation

## References *(optional)*

Fixes 
Connects 


internal/highway/main.go
```diff
@@ -209,6 +209,13 @@ func setupAPI(ctx context.Context, s *HighwayServer) error {
 	s.grpcConn = conn
 	s.grpcClient = v1.NewHighwayClient(conn)
 
+	// Enable CORS
+	// The highway will not be operating under same origin outside of local env
+	s.router.Use(func(c *gin.Context) {
+		c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
```
joshLong145, line 215: do we want to accept all origins? this will be a problem once we deploy highways

joshLong145 (commented):
looks good, might want to make a follow up issue for writing tests once we have the shim in place for highway

internal/highway/main.go
ntindle: Can all those settings just come from cli

joshLong145 (commented):
I think we still need to modify FinishAccessName to return a 'Session' and not the credential standalone. we also need to associate the did with the credential.

PR #310, opened 2022-05-13
Feature/jwt

### LINKED ISSUE #215: Security: Add JWT middleware to http routes
Http routes on Blockchain endpoints are currently unauthenticated. Middleware should be added from Auth0 to support JWT tokens in http headers.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202534370286838) by [Unito](https://www.unito.io)

joshLong145:
Addition of JWT module for highway, which allows for generating and checking for jwt in requests headers. 
Generates new JWT with a configurable secret, expiration, and issuer.
Parses token and generates token, checking for expiration. 
## Changes

- addition of JWT for highway

## Checklist

- [x] Unit tests
- [ ] Documentation

## References *(optional)*
#215 

Fixes 
Connects
closes #215


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/gabriel-vasile/mimetype v1.4.0
 	github.com/gin-gonic/gin v1.7.7
 	github.com/gogo/protobuf v1.3.3
+	github.com/golang-jwt/jwt v3.2.2+incompatible
```
drgomesp, line 14: Is there a specific reason to be using 3.x instead of 4.x here? Just curious, I didn't really go deeply into the differences between the versions.

internal/highway/http/server.go
```diff
@@ -19,12 +19,14 @@ import (
...
 	"github.com/kataras/golog"
-
 	"github.com/sonr-io/sonr/internal/highway/x/ipfs"
+	"github.com/sonr-io/sonr/internal/highway/x/jwt"
 	"github.com/sonr-io/sonr/pkg/client"
 	"github.com/sonr-io/sonr/pkg/config"
 	hn "github.com/sonr-io/sonr/pkg/host"
 	ctv1 "github.com/sonr-io/sonr/x/channel/types"
+	swaggerfiles "github.com/swaggo/files"
+	ginSwagger "github.com/swaggo/gin-swagger"
```
drgomesp: Maybe we could keep package names lowercase? I think that would help avoid confusion and also be a little bit more inline with Go's package/variable naming conventions.

pkg/config/config.go
```diff
@@ -53,6 +54,11 @@ type Config struct {
 	WebAuthNRPIcon        string
 	WebAuthNDebug         bool
 
+	// JWT
+	Secret        string
```
drgomesp: Maybe these should include the JWT prefix? I think from the caller's perspective, `config.Expiration` may be ambiguous. Same goes for Secret.

drgomesp (dismissed):

go.mod
joshLong145: not really, i was testing compatibility and noticed 4x had some major backwards compatibility issues so i went with 3.2.2 I should go back and audit the version 4 release again though. good catch!

internal/highway/http/server.go
joshLong145: good point, i moved these imports without thinking about naming convention. ill make this change.

pkg/config/config.go
joshLong145: fixed

internal/highway/api/middleware.go
```diff
@@ -0,0 +1,52 @@
...
+	Message interface{} `json:"message"`
+}
+
+var middlewareDefs []HighwayMiddleware = nil
+
+func (s *HighwayServer) RegisterMiddleWare() {
+	// Add middleware here to be loaded into Highway's Router
+	middlewareDefs = []HighwayMiddleware{
+		{
+			definition: func(ctx *gin.Context) {
```
mcjcloud: nit: Let's define the actual middleware function separately to clean this up

internal/highway/api/middleware.go
```diff
@@ -0,0 +1,52 @@
...
+				token := ctx.GetHeader("Authorization")
+				if token == "" {
+					ctx.AbortWithStatusJSON(http.StatusBadRequest, UnsignedResponse{
+						Message: "Authorization token not found",
+					})
```
mcjcloud: need `return` after this

internal/highway/api/middleware.go
```diff
@@ -0,0 +1,52 @@
...
+
+				ctx.Next()
+			},
+			disabled: true,
+		},
+	}
+
+	for _, mw := range middlewareDefs {
+		if !mw.disabled {
+			s.Router.Use(mw.definition)
```
mcjcloud, line 32: A JWT may not be needed for every endpoint on highway, so it would make since to apply this on a route-per-route basis.

pkg/jwt/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+var (
+	options JWTOptions = JWTOptions{}
+)
+
+func (opts *JWTOptions) DefaultTestConfig() JWTOptions {
+
+	return JWTOptions{
+		// using rsa pub key as secret
+		secret:        []byte("MIICXAIBAAKBgQCGcXG10nJ5I7OCma6hoH3VMv9OmVV4f8DpSd3Z1I2ud4GpBKFRgCs3AeGLOF87qpGdPh6lCugH3IihKy+vuhISpiploBNyGXEAdHbKyNOCHKvEuOxe9MUEE5oLYvE8qv5LPq5G971PC1HGCwPCglxLXHxzAUnOYsJkFekoWeeBIQIDAQABAoGAMVUPVJiUSL9A73tVCRnLEqBT7pN1OXInZ3MjZPsJwis3+L0qNK3DLbwS9vMIfuxn4jsZI5aM7dWOjRU7uk+csZKPqlvvPyToq8eEBcXWz5R2XpIvzrvucVI4vZMKGgZi8YlzvLnTqdfaGy/AqFDqvmU7fX/aQIu+8bC+DZ1XP8ECQQDJG07O+7hPNtsAAC4lATuIx75SCn+9tHj11I3X7pSV3em+co3XmTYIJsX/G2ehArbUyg5ED8OvjbfxD5qCQT03AkEAqyPm1OhtcI1tq9JGsM1y6av0mM4atEjHvsrxaN0Lt9HyUP8xD52ARxYEilVytYsnArzBGv5sZz7rVpZ2jocgZwJAPjD2tyW7Aqw5H4/utTzjV1JF9gMPK/Bis8surkc2pf4Bagbs/G6B+hVbh5/G9VDsj3OI491oK6MM7jxgEMXyEwJACwZvEnw+wKd7zzvmrfEuW/tl8IomkkK2C4aLctP6s0blM26dPIJLB0lV1YuXrjZetwBt+E03spcNFjDvRlNSNQJBAKd1d0+nm2ruD6vmw7tmJQ/iFIuKRMtcTKd2nVhO7TCw7ORM4tH56NWj74uioXQ1YmfFHvwZY3O0woLTa91scYk="),
```
mcjcloud, line 21: This needs to be an env variable and an actual secret. It should only be known by highways

pkg/jwt/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func DefaultNew() JWT {
+	return JWT{
+		options: options.DefaultTestConfig(),
+	}
+}
+
+/*
+
```
mcjcloud: empty comment

pkg/jwt/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+ */
+func (j *JWT) Generate(doc did.Document) (string, error) {
+	if doc == nil {
```
mcjcloud, line 39: Since you're not using a pointer argument, this will never be true (only pointers can be nil). Consider instead checking that the fields in the struct are valid.

pkg/jwt/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+		return "", errors.New("token type is not JWT")
+	}
+	// Sign and get the complete encoded token as a string using the secret
+	tokenString, err := token.SignedString(j.options.secret)
+
+	return tokenString, err
+}
+
+/*
+
```
mcjcloud: empty comment

pkg/jwt/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+ */
+func (j *JWT) Parse(token string) (*jwt.Token, error) {
+	parsed, err := jwt.ParseWithClaims(token, &jwt.StandardClaims{}, func(token *jwt.Token) (interface{}, error) {
+		return j.options.secret, nil
+	})
+
+	if parsed.Valid {
```
mcjcloud: Did you intend to do something here? If not you can just remove it.

pkg/jwt/jwt_test.go
```diff
@@ -0,0 +1,169 @@
...
+	priv := secp256k1.GenPrivKey()
+	pub := priv.PubKey()
+	vm, _ := did.NewVerificationMethod(
+		*didUrl,
+		ssi.ECDSASECP256K1VerificationKey2019,
+		*didController,
+		pub,
+	)
+
+	thing := func() *did.Document {
```
mcjcloud: nit: rename to something more specific like `createDocWithAuthMethod`

pkg/jwt/jwt_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+)
+
+func Test_JWT(t *testing.T) {
+	root := "123456"
+
+	didController, err := did.ParseDID(fmt.Sprintf("did:snr:%s#test", root))
+	if err != nil {
+		t.Errorf("Error while creating test controller %s", err)
```
mcjcloud, line 21: There's also `assert.NoError`

pkg/jwt/jwt_test.go
```diff
@@ -0,0 +1,169 @@
...
+		tokenObj, error := jwt.Parse(token)
+		if error != nil {
+			t.Errorf("Error while generating token %s", err)
+		}
+		claims, ok := GetClaims(tokenObj)
+		if !ok {
+			t.Error("Unable to parse claims from token")
+		}
+
+		assert.NotNil(t, claims.Issuer)
```
mcjcloud, line 106: Add a check to compare the DIDs

pkg/jwt/utils.go
```diff
@@ -0,0 +1,27 @@
...
+	jwtToken := strings.Split(header, " ")
+	if len(jwtToken) != 2 {
+		return "", errors.New("incorrectly formatted authorization header")
+	}
+
+	return jwtToken[1], nil
+}
+
+func GetClaims(token *jwt.Token) (*jwt.StandardClaims, bool) {
+	claims, ok := token.Claims.(*jwt.StandardClaims)
```
mcjcloud, line 24: nit: This could just be `return token.Claims.(*jwt.StandardClaims)`.

mcjcloud (dismissed):

internal/highway/api/middleware.go
joshLong145: can we get a better definition on separately?

internal/highway/api/middleware.go
joshLong145: crap, good catch lol

internal/highway/api/middleware.go
joshLong145: i dont know if gin has a notion of contextual middleware but i can check, if not then it has to be explicitly implemented in every route that requires it.

pkg/jwt/config.go
joshLong145: yes, this was just left over from testing, ill add it as env

pkg/jwt/jwt_test.go
joshLong145: wow i thought i had that in this test, thanks!

internal/highway/api/middleware.go
drgomesp: I believe @mcjcloud means (pleas correct me if I'm wrong) we could define the authorization middleware in its own separate file as an isolated function, and just reference it from here.

internal/highway/api/middleware.go
drgomesp: There is a way to set a middleware on a per-route basis (https://github.com/gin-gonic/gin#using-middleware), you'd just have to pass the middleware when defining the route itself.

pkg/jwt/jwt.go
drgomesp: You can also check for the zero-value, which makes sense here:

```go
// if is zero-value
if doc == did.Document{} {
}
```

Although I'd really prefer that this becomes a pointer too.

internal/highway/api/server.go
```diff
@@ -45,8 +47,9 @@ type HighwayServer struct {
 	Config *config.Config
 
 	// Clients
-	Host   hn.SonrHost
-	Cosmos *client.Cosmos
+	Host     hn.SonrHost
+	Cosmos   *client.Cosmos
+	JwtToken *jwt.JWT
```
drgomesp: Could we name this `JWTToken` here instead for consistency purposes?

pkg/config/config.go
```diff
@@ -46,6 +47,11 @@ type Config struct {
 	HighwayGRPCEndpoint string
 	HighwayHTTPEndpoint string
 
+	// JWT
+	JWT_Secret        string
```
drgomesp: These could be named without underscores (`JWTSecret`, `JWTSigningMethod`, etc) for more consistency with Go standards.

pkg/jwt/jwt.go
```diff
@@ -0,0 +1,82 @@
...
+	options JWTOptions
+}
+
+func New(ctx context.Context, hn host.SonrHost) JWT {
+	options = JWTOptions{}
+	options.secret = []byte(hn.Config().JWT_Secret)
+	options.singingMethod = hn.Config().JWT_SigningMethod
+	options.ttl = hn.Config().JWT_Expiration
+
+	return JWT{
```
drgomesp: Can we make this a pointer?

pkg/jwt/jwt_test.go
```diff
@@ -0,0 +1,169 @@
...
+		jwt := DefaultNew()
+		assert.NotNil(t, jwt)
+		assert.NotNil(t, jwt.options)
+	})
+
+	t.Run("Should generate JWT from did uri", func(t *testing.T) {
+		doc := thing()
+		jwt := DefaultNew()
+		token, err := jwt.Generate(*doc)
+		if err != nil {
```
drgomesp: You could use `assert.NoError` here and on the rest of error checks.

pkg/jwt/middleware.go
```diff
@@ -0,0 +1,23 @@
+package jwt
+
+import (
+	"errors"
+)
+
+type JWTParserMiddleware = func() error
```
drgomesp, line 7: For type definitions, it's more appropriate to remove the = sign. In this case, this is actually an alias and not a type definition.

internal/highway/api/server.go
```diff
@@ -45,8 +47,9 @@ type HighwayServer struct {
 	Config *config.Config
 
 	// Clients
-	Host   hn.SonrHost
-	Cosmos *client.Cosmos
+	Host     hn.SonrHost
+	Cosmos   *client.Cosmos
+	JWTToken *jwt.JWT
```
mcjcloud, line 52: nit: JWTToken is redundant, maybe `TokenClient`?

pkg/jwt/middleware.go
mcjcloud: +1

drgomesp (approved):

drgomesp:
Just needs a rebase, but looks good to merge!

prdlk (approved):

PR #338, opened 2022-05-25
Feature/motor state machine

### LINKED ISSUE #286: [LOGIC]: Motor State Methods
**Overview**
Underlying motor node state management methods into frameworks

**Involved Methods**

* [ ] `Start()`
* [ ] `Stop()`
* [ ] `Pause()`
* [ ] `Resume()`
* [ ] `Status()`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202534120946035) by [Unito](https://www.unito.io)

joshLong145:
Addition of motor state methods for Host interface and Host Implementation
Migration of HostStatus and implementation of status into separate file "state_machine"
Addition of transition mappings

closes #286

## Changes

* Additional state methods on Host Interface
* Migration of state into a "State Machine" primitive implementation

Below is a diagram outlining the implemented state transitions
image

## API Updates

Start()
Stop()
Pause()
Resume()
Status()

## Checklist

* [x] Unit tests
* [ ] Documentation

## References *(optional)*

Fixes
Connects


pkg/host/utils.go
```diff
@@ -0,0 +1,18 @@
+package host
```
prdlk: Adding config to host breaks the motor build as viper is not compatible with WASM and Mobile

prdlk (changes requested):

pkg/host/utils.go
joshLong145: good call, i rembered that and took out the helpers i wrote using some env vars but didn't remove the file entirely. ill do that.

prdlk (dismissed):

drgomesp (dismissed):

drgomesp (dismissed):

pkg/host/config.go
```diff
@@ -94,112 +84,20 @@ type SonrHost interface {
...
-	case Status_READY:
-		return "READY"
-	case Status_FAIL:
-		return "FAIL"
-	case Status_CLOSED:
-		return "CLOSED"
-	}
-	return "UNKNOWN"
-}
+	VerifyData(data []byte, signature []byte, peerId peer.ID, pubKeyData []byte) error
```
mcjcloud, line 87: The previous signature for this method is clearer

pkg/host/config.go
```diff
@@ -94,112 +84,20 @@ type SonrHost interface {
...
 
-// SetStatus sets the host status and emits the event
-func (h *hostImpl) SetStatus(s HostStatus) {
-	// Check if status is changed
-	if h.status == s {
-		return
-	}
+	// Close closes the node
+	Close()
```
mcjcloud: I think most of these should return `error`, because their current implementations just suppress any errors that pop up. Might make debugging harder.

pkg/host/host.go
```diff
@@ -213,20 +214,110 @@ func (hn *hostImpl) createDHTDiscovery(c *config.Config) error {
...
+
+/*
+	Starts the libp2p host, dhcp, and sets the host status to ready
+*/
+func (hn *hostImpl) Start() {
+	// Create Connection Manager
+	c := hn.config
+	cnnmgr, err := cmgr.NewConnManager(c.Libp2pLowWater, c.Libp2pHighWater)
+	if err != nil {
+		return
```
mcjcloud: set status as fail?

pkg/host/host.go
```diff
@@ -213,20 +214,110 @@ func (hn *hostImpl) createDHTDiscovery(c *config.Config) error {
...
+	hn.host, err = libp2p.New(
+		libp2p.Identity(hn.privKey),
+		libp2p.ConnectionManager(cnnmgr),
+		libp2p.DefaultListenAddrs,
+		libp2p.Routing(hn.Router),
+		libp2p.EnableAutoRelay(),
+	)
+
+	if err != nil {
+		hn.fsm.SetStatus(Status_FAIL)
```
mcjcloud: return?

pkg/host/host.go
```diff
@@ -213,20 +214,110 @@ func (hn *hostImpl) createDHTDiscovery(c *config.Config) error {
...
+	}
+	hn.fsm.SetStatus(Status_CONNECTING)
+
+	// Connect to Bootstrap Nodes
+	for _, pi := range c.Libp2pBootstrapPeers {
+		if err := hn.Connect(pi); err != nil {
+			continue
+		} else {
+			hn.fsm.SetStatus(Status_FAIL)
```
mcjcloud: You're setting status to FAIL if there's no error here. Is that right?

pkg/host/host.go
```diff
@@ -213,20 +214,110 @@ func (hn *hostImpl) createDHTDiscovery(c *config.Config) error {
...
+/*
+	Stops the libp2p host, dhcp, and sets the host status to IDLE
+*/
+func (hn *hostImpl) Stop() {
+	err := hn.host.Close()
+	if err != nil {
+		hn.fsm.SetStatus(Status_FAIL)
+		return
+	}
+	defer hn.Pause()
```
mcjcloud: This doesn't really need to be deferred since it's at the end of the function

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+
+// HostStatus is the status of the host
+type HostStatus string
+
+// Equals returns true if given SNRHostStatus matches this one
+func (s HostStatus) Equals(other HostStatus) bool {
+	return s == other
+}
+
+// IsNotIdle returns true if the SNRHostStatus != Status_IDLE
```
mcjcloud: Update comment `IsIdle`

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+func (s HostStatus) IsIdle() bool {
+	return s != Status_IDLE
```
mcjcloud: Function is called `IsIdle` but is returning "is not idle"

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+func (s HostStatus) IsFail() bool {
+	return s == Status_FAIL
+}
+
+// IsClosed returns true if the SNRHostStatus == Status_CLOSED
+func (s HostStatus) IsClosed() bool {
+	return s == Status_CLOSED
+}
+
+func (sm *SFSM) GetCurrent() string {
```
mcjcloud: nit: This doesn't need to be a pointer.

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+	return string(sm.Current)
+}
+
+type SFSM struct {
+	Current      HostStatus
+	Chn          chan bool
```
mcjcloud: What is this for? It can probably have a better name

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+func (s HostStatus) IsClosed() bool {
+	return s == Status_CLOSED
+}
+
+func (sm *SFSM) GetCurrent() string {
+	return string(sm.Current)
+}
+
+type SFSM struct {
+	Current      HostStatus
```
mcjcloud: `Status` or `CurrentStatus` would be more clear

pkg/host/state_machine.go
```diff
@@ -0,0 +1,133 @@
...
+	return &SFSM{
+		States:       &states,
+		StateMapping: &STATE_MAPPINGS,
+		Current:      Status_IDLE,
+	}
+
+}
+
+// SetStatus sets the host status and emits the event
+func (fsm *SFSM) SetStatus(s HostStatus) {
```
mcjcloud: Should there be an error if it fails to transition states?

mcjcloud: nit: It's a state machine, but `Status` is being used a lot instead of `State`.

mcjcloud (changes requested):

pkg/host/state_machine.go
joshLong145: I kept the original wording of the interface, Perhaps we should use `State` instead of `Status` but it would change the methods on the host interface

joshLong145: Agreed, ill change this to return an error to the caller.

pkg/host/state_machine.go
joshLong145: yea i should rename Chn to me more explicit in its usage.

pkg/host/state_machine.go
joshLong145: fair, the idea was to save a bit of memory not copying the value every time. but im not totally sure if the type declaration creates a struct wrapper or the compiler swaps it out for the primitive type. either way i think it does save us a bit of memory using a pointer here.

pkg/host/config.go
joshLong145: thats fair, i didn't refactor the interfaces in this implementation. perhaps that is a separate task?

pkg/host/host.go
joshLong145: Do you mean set the status to `Fail` in the scope of the error check?

pkg/host/host.go
joshLong145: return in the scope of the error case?

pkg/host/host.go
joshLong145: i think this sets status to fail in the case of an error on connection

pkg/host/state_machine.go
joshLong145: Good catch

pkg/host/host.go
mcjcloud: On line 274 you have `err != nil`, so you're continuing if there's an error and setting status to fail if there's not.

mcjcloud (approved):

PR #339, opened 2022-05-26
Fix queries, these have been validated as functional by client impl

prdlk:
I was running the blockchain with the client impl and realized that the query methods had a little glitch. This is a quick refactor and has been tested with HTTP, gRPC, and Dart Client.

https://user-images.githubusercontent.com/3804558/170404158-adbc545c-d396-4d74-ac09-cb2ca53c611c.mp4


prdlk:
@ntindle We would be ready to deploy v0.1.0 devnet after this is merged

ntindle (dismissed):

x/registry/keeper/grpc_query_who_is.go
```diff
@@ -47,11 +47,49 @@ func (k Keeper) WhoIs(c context.Context, req *types.QueryWhoIsRequest) (*types.Q
...
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	val, found := k.FindWhoIsByAlias(
+		ctx,
+		req.Alias,
+	)
+	if !found {
+		return nil, status.Error(codes.InvalidArgument, "not found")
```
mcjcloud: `InvalidArgument` should be `NotFound`

x/registry/keeper/grpc_query_who_is.go
```diff
@@ -47,11 +47,49 @@ func (k Keeper) WhoIs(c context.Context, req *types.QueryWhoIsRequest) (*types.Q
...
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	val, found := k.FindWhoIsByController(
+		ctx,
+		req.Controller,
+	)
+	if !found {
+		return nil, status.Error(codes.InvalidArgument, "not found")
```
mcjcloud: Same as above, `NotFound`

x/registry/keeper/msg_server_who_is.go
```diff
@@ -86,7 +86,7 @@ func (k msgServer) UpdateWhoIs(goCtx context.Context, msg *types.MsgUpdateWhoIs)
 			val.AddAlsoKnownAs(a, false)
 		}
 	}
-	val.Controllers = doc.ControllersAsString()
+	val.Controllers = doc.GetController()
```
mcjcloud, line 89: This method should be named `GetControllers`. I know it's not part of this PR but we should at least put an issue or just change it here.

mcjcloud (dismissed):

mcjcloud:
Why were the alias methods removed?

x/registry/keeper/msg_server_who_is.go
prdlk: If anything WhoIs should be renamed to `Controller` that is the compliant property with the DID W3 Specification

ntindle:
> Why were the alias methods removed?

Weren't the files just merged?

x/registry/keeper/msg_server_who_is.go
joshLong145: This is out of scope.

joshLong145 (approved):

ntindle (approved):

PR #360, opened 2022-06-06
Remove/ibc go

prdlk:
## Changes

* Removes IBC-Go dependency from go.mod and app.go


go.mod
```diff
@@ -26,7 +25,7 @@ require (
 	github.com/libp2p/go-libp2p-connmgr v0.3.1
 	github.com/libp2p/go-libp2p-core v0.13.0
 	github.com/libp2p/go-libp2p-discovery v0.6.0
-	github.com/libp2p/go-libp2p-http v0.2.1
+	github.com/libp2p/go-libp2p-http v0.2.0
```
joshLong145, line 28: is there a reason to downgrade the minor in this context?

prdlk: We can bind motor to wasm in this version

PR #378, opened 2022-06-13
Feature/jwk interface update

mcjcloud:
Update JWX interface and default implementation to keep track of encryption/signing algorithms.

## Changes

- Created JWX interface
- Created `jwxImpl` struct (implements the interface and keeps track of algs)
- Switched jwx library from v1 to v2


pkg/jwx/jwk_test.go
```diff
@@ -16,8 +16,9 @@ func Test_JWK(t *testing.T) {
 		if err != nil {
 			t.Errorf("Error while convering credential to public key %s", err)
 		}
-		pk := key.Public()
-		jwk, err := CreateJWKForEnc(pk)
+
+		x := New(key.Public())
```
joshLong145, line 20: why do we need to wrap the pk with this "New"

joshLong145 (approved):
approving but left a question

PR #380, opened 2022-06-14
Persistent Store

drgomesp:
Persistence is now happening through the node's filesystem.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202529369355809) by [Unito](https://www.unito.io)
┆Attachments: #380 Persistent Store

### DIFF
--- a/internal/highway/x/ipfs/protocol.go
+++ b/internal/highway/x/ipfs/protocol.go
@@ -25,7 +25,9 @@ type Protocol struct {
 }
 
 // New creates a new Protocol instance with Host Implementation
-func New(ctx context.Context, ds *store.Store, host host.SonrHost) (*Protocol, error) {
+func New(ctx context.Context, host host.SonrHost) (*Protocol, error) {
+	ds := store.New("tmp")
+
 	// Create IPFS Peer
 	peer, err := ipfslite.New(ctx, ds, host.Host(), host.Routing(), nil)
 	if err != nil {
--- a/internal/highway/x/ipfs/protocol.go
+++ b/internal/highway/x/ipfs/protocol.go
@@ -26,8 +26,8 @@ type IPFSProtocol struct {
 // New creates a new IPFSProtocol instance with Host Implementation
 func New(ctx context.Context, host host.SonrHost) (*IPFSProtocol, error) {
 	// Create IPFS Peer
-	ds := NewStore()
-	ipfsLite, err := ipfslite.New(ctx, ds.Batching(), host.Host(), host.Routing(), nil)
+	ds := NewStore("tmp")
+	ipfsLite, err := ipfslite.New(ctx, ds, host.Host(), host.Routing(), nil)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/highway/x/ipfs/protocol.go
+++ b/internal/highway/x/ipfs/protocol.go
@@ -4,16 +4,17 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/sonr-io/sonr/internal/highway/x/store"
-	"io/ioutil"
-
 	ipfslite "github.com/hsanjuan/ipfs-lite"
 	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
 	"github.com/ipld/go-ipld-prime/codec/dagjson"
 	"github.com/ipld/go-ipld-prime/datamodel"
 	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	"github.com/sonr-io/sonr/internal/highway/x/store"
 	"github.com/sonr-io/sonr/pkg/host"
 	st "github.com/sonr-io/sonr/x/schema/types"
+	"io/ioutil"
+	"log"
 )
 
 // Protocol leverages the IPFSLite library to provide simple file operations.
@@ -56,6 +57,25 @@ func (i *Protocol) GetData(cid string) ([]byte, error) {
 	c, err := DecodeCIDFromString(cid)
 	if err != nil {
 		return nil, err
+	}
+
+	ctx := context.TODO()
+	key := datastore.NewKey(cid)
+
+	exists, err := i.store.Has(ctx, key)
+	if err != nil {
+		// log error and continue
+		log.Println(err)
+	}
+
+	if exists {
+		data, err := i.store.Get(ctx, key)
+		if err != nil {
+			// log error and continue
+			log.Println(err)
+		}
+
+		return data, nil
 	}
 
 	// Get the file from IPFS
@@ -102,6 +122,12 @@ func (i *Protocol) PutData(data []byte) (*cid.Cid, error) {
 
 	// Get Back the CID
 	c := nd.Cid()
+
+	// Use the CID to store the file on local fs.
+	if err := i.store.Put(context.TODO(), datastore.NewKey(c.String()), data); err != nil {
+		return nil, err
+	}
+
 	return &c, nil
 }
 
@@ -160,5 +186,22 @@ func (i *Protocol) RemoveFile(cidstr string) error {
 	if err != nil {
 		return err
 	}
+
+	ctx := context.TODO()
+	key := datastore.NewKey(cid.String())
+
+	exists, err := i.store.Has(ctx, key)
+	if err != nil {
+		// log error and continue
+		log.Println(err)
+	}
+
+	if exists {
+		if err := i.store.Delete(ctx, key); err != nil {
+			// log error and continue
+			log.Println(err)
+		}
+	}
+
 	return i.Peer.Remove(i.ctx, cid)
 }
--- a/internal/highway/x/ipfs/protocol.go
+++ b/internal/highway/x/ipfs/protocol.go
@@ -19,14 +19,14 @@ import (
 type IPFSProtocol struct {
 	ctx       context.Context
 	node      host.SonrHost
-	dataStore *MemoryStore
+	dataStore *Store
 	*ipfslite.Peer
 }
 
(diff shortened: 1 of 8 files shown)


internal/highway/x/store/store.go
```diff
@@ -0,0 +1,116 @@
...
+			result.Entry.Key = key.String()
+			if !q.KeysOnly {
+				result.Entry.Value, result.Error = s.Get(ctx, key)
+			}
+			results <- result
+		}
+		return nil
+	}
+
+	go func() {
```
joshLong145: I'm thinking this would be better as a deferred. Reason being it will be easier to reason in the scenario where multiple requests are invoking this method. However my thought process could be flawed here.

joshLong145: my thinking is that we should close after the function returns in the same routine context, not spawn another routine where we could have many of these running at the same time under high load.

drgomesp: @joshLong145 regarding the first comment, do you mean in case multiple requests are coming in? In that case, every `go func()` is dispatched in its own context, so I don't really see where this could be a problem, but maybe I'm missing something.

drgomesp: I see what you mean. Will change it and see if that's better.

internal/highway/x/ipfs/protocol.go
```diff
@@ -4,39 +4,42 @@ import (
...
+	host  host.SonrHost
+	store *store.Store
 	*ipfslite.Peer
 }
 
-// New creates a new IPFSProtocol instance with Host Implementation
-func New(ctx context.Context, host host.SonrHost) (*IPFSProtocol, error) {
+// New creates a new Protocol instance with Host Implementation
+func New(ctx context.Context, host host.SonrHost) (*Protocol, error) {
+	ds := store.New("tmp")
```
mcjcloud: Use `os.TempDir()` instead of "tmp" to support more platforms.

internal/highway/x/ipfs/protocol.go
```diff
@@ -49,13 +52,32 @@ func DecodeCIDFromString(s string) (cid.Cid, error) {
...
 // GetData returns a file from IPFS.
-func (i *IPFSProtocol) GetData(cid string) ([]byte, error) {
+func (i *Protocol) GetData(cid string) ([]byte, error) {
 	// Decode CID from String
 	c, err := DecodeCIDFromString(cid)
 	if err != nil {
 		return nil, err
 	}
 
+	ctx := context.TODO()
```
mcjcloud: Use `context.Background()` for a new context

internal/highway/x/ipfs/protocol.go
```diff
@@ -49,13 +52,32 @@ func DecodeCIDFromString(s string) (cid.Cid, error) {
...
+	key := datastore.NewKey(cid)
+
+	exists, err := i.store.Has(ctx, key)
+	if err != nil {
+		// log error and continue
+		log.Println(err)
```
mcjcloud: Is it possible for exists to be `true` and err to not be nil?

If not I would suggest returning the error here.

internal/highway/x/ipfs/protocol.go
```diff
@@ -49,13 +52,32 @@ func DecodeCIDFromString(s string) (cid.Cid, error) {
...
+	}
+
+	if exists {
+		data, err := i.store.Get(ctx, key)
+		if err != nil {
+			// log error and continue
+			log.Println(err)
```
mcjcloud: Same comment as above. We should probably just return error here.

internal/highway/x/ipfs/protocol.go
```diff
@@ -100,11 +122,17 @@ func (i *IPFSProtocol) PutData(data []byte) (*cid.Cid, error) {
 
 	// Get Back the CID
 	c := nd.Cid()
+
+	// Use the CID to store the file on local fs.
+	if err := i.store.Put(context.TODO(), datastore.NewKey(c.String()), data); err != nil {
```
mcjcloud: use `context.Background()`

internal/highway/x/ipfs/protocol.go
```diff
@@ -153,10 +181,27 @@ func (i *IPFSProtocol) PutObjectSchema(doc *st.SchemaDefinition) (*cid.Cid, erro
...
 
 // RemoveFile removes a file from IPFS.
-func (i *IPFSProtocol) RemoveFile(cidstr string) error {
+func (i *Protocol) RemoveFile(cidstr string) error {
 	cid, err := DecodeCIDFromString(cidstr)
 	if err != nil {
 		return err
 	}
+
+	ctx := context.TODO()
```
mcjcloud: `context.Background()`

internal/highway/x/ipfs/protocol.go
```diff
@@ -153,10 +181,27 @@ func (i *IPFSProtocol) PutObjectSchema(doc *st.SchemaDefinition) (*cid.Cid, erro
...
+	exists, err := i.store.Has(ctx, key)
+	if err != nil {
+		// log error and continue
+		log.Println(err)
+	}
+
+	if exists {
+		if err := i.store.Delete(ctx, key); err != nil {
+			// log error and continue
+			log.Println(err)
```
mcjcloud, line 199: same comments about returning errors from above.

internal/highway/x/store/store.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *Store) Has(ctx context.Context, key datastore.Key) (exists bool, err error) {
+	return s.dataStore.Has(ctx, key.String())
+}
+
+func (s *Store) GetSize(ctx context.Context, key datastore.Key) (size int, err error) {
+	return 0, errors.New("TODO")
+}
+
+func (s *Store) Query(ctx context.Context, q query.Query) (query.Results, error) {
+	log.Printf("querying for %+v", q)
```
mcjcloud: nit: include newline in print

mcjcloud (changes requested):
Add unit tests

internal/highway/x/ipfs/protocol.go
drgomesp: SInce this is somewhat like a fallback strategy for persisting, I thought we shouldn't return the error and potentially break the flow, if we can log it and proceed to save it in IPFS. However that's definitely something that can be discussed or changed, just trying to explain what was the process to arrive at this code.

internal/highway/x/ipfs/protocol.go
joshLong145: agreed

internal/highway/x/ipfs/store.go
```diff
@@ -1,38 +0,0 @@
...
-	"context"
-
-	"github.com/ipfs/go-datastore"
-
-	ipfslite "github.com/hsanjuan/ipfs-lite"
-	"github.com/ipld/go-ipld-prime/storage"
-)
-
-// MemoryStore is a datastore.Batching implementation that stores data in memory.
-type MemoryStore struct {
```
joshLong145, line 13: I feel we should still keep the mem store and have both a in memory store and an "fs" storage

internal/highway/x/store/store.go
joshLong145: not we should not log in these implementations as the logger is not compatible with all compilation targets.

internal/highway/x/store/store.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *Store) Close() error {
+	return errors.New("TODO")
+}
+
+func (s *Store) DiskUsage(ctx context.Context) (uint64, error) {
+	return 0, errors.New("TODO")
+}
+
+func (s *Store) Batch(ctx context.Context) (datastore.Batch, error) {
+	return nil, errors.New("TODO")
```
joshLong145: Batching needs to be implemented as it is the default strategy  for persisting content to my understanding of IPFS light

joshLong145 (changes requested):
see comments inline

internal/highway/x/ipfs/protocol.go
drgomesp: I was actually aiming for a root path's `tmp` dir and not the system one's, but system tmp seems better indeed.

internal/highway/x/ipfs/protocol.go
drgomesp: The reason I used TODO here is because I assumed the context should be coming from outside, and be passed to the protocol functions. I will make it background now, but I think we should rethink these.

prdlk (commented):
Confused about this Driver directory can you provide some context?

drgomesp:
@prnk28 the idea here is to have the ability to use the filesystem as the actual persistence layer and have the memory store (as we used to have) as a cache layer. This exposes the need to actually abstract away these into what I'm calling the drivers (`driver.FileSystem` and `driver.Memory`) here, which are essentially where the storage-specific code goes. This means we would introduce a more high-level and generic API which we are calling here the `store.Store`. 

After discussing this with @joshLong145 however, we agreed that it doesn't make sense anymore to have a fs layer since this is mostly going to be replaced by the IPFS layer, sort of virtualizing the FS in a way. In this case, we are going to remove the protocol implementation (`ipfs.Protocol`) and create a new, more generic API which goes in `pkg/{something}/{something}.go`.

This is what I'm currently working on, which is designing this interface and find a way to have the ability to abstract away the inner layers (IPFS, FileCoin, whatever else we decide we can use) under a storage mechanism which is generic enough, but that considers the decentralized nature of the storage (having DAGs, CIDs, etc).

prdlk:
Okay this is a blocker for testing objects getting stored/received persistently - let's try to merge this ASAP

PR #384, opened 2022-06-17
Feature/create schema

joshLong145:
- Functionality to create `Schemas` which persist to storage mechanisms

## Changes
- Addition of the following proto definitions
    - WhatIs
    - SchemaReference
    - SchemaDefinition

## API Updates
- /v1/schema/create (HTTP Highway)
### Deprecations *(required)*
- None

## Checklist

- [x] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 
closes #374 


x/schema/keeper/msg_server_schema.go
```diff
@@ -3,13 +3,69 @@ package keeper
...
+	}
+
+	var schema = types.SchemaReference{
+		Label: msg.Label,
+		Did:   what_is_did.String(),
+		Cid:   msg.Cid,
+	}
+
+	var whatIs = types.WhatIs{
+		Creator:   creator_did.GetID().String(),
```
mcjcloud: Seems like you're creating a did document in `creator_did` only to take the DID string from it (which you already had)

x/schema/keeper/schema.go
```diff
@@ -0,0 +1,94 @@
...
+	"encoding/binary"
+
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/google/uuid"
+	"github.com/sonr-io/sonr/x/schema/types"
+)
+
+func (k Keeper) GenerateKeyForDID() string {
+	id := uuid.New()
```
mcjcloud: nit: `return uuid.New().String()`

x/schema/keeper/schema.go
```diff
@@ -0,0 +1,94 @@
...
+
+// GetSchemaFromCreator returns a WhoIs whos DIDDocument contains the given controller
+func (k Keeper) GetWhatIsFromCreator(ctx sdk.Context, creator string) (val []types.WhatIs, found bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.SchemaKeyPrefix))
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
+	defer iterator.Close()
+	var vals []types.WhatIs = make([]types.WhatIs, 0)
+	for ; iterator.Valid(); iterator.Next() {
+		var instance types.WhatIs
+		k.cdc.MustUnmarshal(iterator.Value(), &instance)
```
mcjcloud: I don't think we should panic if this fails. We won't be able to return a valid error response

x/schema/module.go
```diff
@@ -77,7 +79,7 @@ func (AppModuleBasic) RegisterRESTRoutes(clientCtx client.Context, rtr *mux.Rout
 
 // RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the module.
 func (AppModuleBasic) RegisterGRPCGatewayRoutes(clientCtx client.Context, mux *runtime.ServeMux) {
-	// this line is used by starport scaffolding # 2
```
mcjcloud, line 80: Don't remove this comment

mcjcloud (approved):

x/schema/keeper/msg_server_schema.go
prdlk: Yes better to just create a method in `pkg/did` which parses any arbitrary snrAddress and converts it to the `did:snr:` equivalent

x/schema/keeper/schema.go
prdlk: @ntindle Thoughts?

prdlk (dismissed):

x/schema/keeper/msg_server_schema.go
joshLong145: good point, i'll make that change.

prdlk (approved):

PR #396, opened 2022-06-24
Refactor/schema

joshLong145:
Migration of IPFS schema record persistence from highway rest endpoint to x/schema module (blockchain)

Addition of IPFS content lookup in x/schema module Query (blockchain)
changes to schema message proto definitions

## Changes

* migration of IPFS schema persistence from highway rest endpoint to x/schema `CreateSchema` on chain.
* Addition of IPFS content lookup in x/schema query method `QuerySchema`.

New Proto Definitions for x/Schema described here:
**Message type definitions**

```
message MsgCreateSchema {
  // definition of schema to be persisted
  SchemaDefinition definition = 1;
}

message MsgCreateSchemaResponse {
  // Status code of the response
  int32 code = 1;

  string message = 2;

  // WhatIs object created on chain
  WhatIs what_is = 3;
}

message MsgDeprecateSchema {
  string creator = 1;
  string did = 2;
}

message MsgDeprecateSchemaResponse {
  int32 code = 1;
  string message = 2;
}
```

**Query Definitions**

```
message QuerySchemaRequest {
  string creator = 1;
  string did = 2;
}

message QuerySchemaResponse {
  // Schema Representation
  WhatIs whatIs = 1;

  // Definition of schema
  SchemaDefinition definition = 2;
}
```

**Schema definitions**

```
syntax = "proto3";

package sonrio.sonr.schema;

option go_package = "github.com/sonr-io/sonr/x/schema/types";

// Schema defines the shapes of schemas on Sonr
message SchemaReference {
  // the DID for this schema
  string did = 1;

  // an alternative reference point
  string label = 2;

  // a reference to information stored within an IPFS node.
  string cid = 3;
}

message SchemaDefinition {
  // the DID for this schema
  string creator = 1;

  // an alternative reference point
  string label = 2;

  // the properties of this schema
  map<string, SchemaKind> fields = 3;
}

// Represents the types of fields a schema can have
enum SchemaKind {
  INVALID = 0;
  MAP = 1;
  LIST = 2;
  UNIT = 3;
  BOOL = 4;
  INT = 5;
  FLOAT = 6;
  STRING = 7;
  BYTES = 8;
  LINK = 9;
  STRUCT = 10;
  UNION = 11;
  ENUM = 12;
  ANY = 13;
}
```

**WhatIs definition**

```
message WhatIs {
  // DID is the DID of the object
  string did = 1;

  // Object_doc is the object document
  SchemaReference schema = 2;

  // Creator is the DID of the creator
  string creator = 3;

  // Timestamp is the time of the last update of the DID Document
  int64 timestamp = 4;

  // IsActive is the status of the DID Document
  bool is_active = 5;
}
```

## API Updates

QuerySchema now looks up pinned IPFS content from

## Checklist

* [] Unit tests

* [ ] Documentation

## References *(optional)*

Fixes
Connects


web/wallet/src/store/generated/package.json
```diff
@@ -1,17 +0,0 @@
-{
```
ntindle, line 1: Why deleted?

proto/schema/v1/query.proto
```diff
@@ -40,7 +41,11 @@ message QuerySchemaRequest {
 }
 
 message QuerySchemaResponse {
-  WhatIs schema = 1;
+  // Schema Repersentation
+  WhatIs whatIs = 1;
```
mcjcloud, line 54: I think it's unnecessary to include both of these. Just the definition should be fine

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
 	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/google/uuid"
+	shell "github.com/ipfs/go-ipfs-api"
 	"github.com/sonr-io/sonr/x/schema/types"
 )
 
+const (
+	URL_PERSISTENCE_READ = "ipfs.Sonr.ws"
+)
```
ntindle: if not a ton of effort, pull from env?

proto/schema/v1/tx.proto
```diff
@@ -15,14 +15,17 @@ service Msg {
 }
 
 message MsgCreateSchema {
-  string creator = 1;
-  string label = 2;
-  string cid = 3;
+  // definition of schema to be persisted
```
mcjcloud, line 18: This will still need the creator for now

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
 
 	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/google/uuid"
+	shell "github.com/ipfs/go-ipfs-api"
 	"github.com/sonr-io/sonr/x/schema/types"
 )
 
+const (
+	URL_PERSISTENCE_READ = "ipfs.Sonr.ws"
```
mcjcloud: nit: Sonr should be lowercase

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	defer os.Remove(out_path)
+
+	return nil
+}
```
ntindle: Not in love with this downloading then reading the file, would prefer to do this in memory, but there are some tradeoffs there. Want thoughts on that

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	if err = json.Unmarshal(resp, &content); err != nil {
+		return err
+	}
+
+	if err != nil {
```
mcjcloud, line 48: You've already done this check.

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
+}
+
+func (k Keeper) PinContent(payload interface{}) (string, error) {
+	b, err := json.Marshal(payload)
+	if err != nil {
+		return "", err
+	}
+	return ipfs_inter.Add(bytes.NewReader(b))
+}
+
```
ntindle, line 62: this probably won't work as the endpoint for uploading is different from downloading. One of these two functions will probaly break.

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,62 @@
...
+	if err = json.Unmarshal(resp, &content); err != nil {
+		return err
+	}
+
+	if err != nil {
+		return err
+	}
+
+	defer os.Remove(out_path)
```
mcjcloud: nit: This should be right after `out_path` is declared

mcjcloud (changes requested):

ntindle (changes requested):
take a look at the comment on ipfs endpoints diffing

x/schema/keeper/schema.go
prdlk: Memory is better

proto/schema/v1/query.proto
prdlk: What if we need to verify the schema is registered for current app? And situation where App B wants to use schemas from App A? I feel like keeping WhatIs will help alleviate potential unaccounted for issues.
we can create a separate endpoint for querying the `WhatIs` and it will be faster without the added content lookup.

proto/schema/v1/tx.proto
prdlk: +1

x/schema/keeper/schema.go
prdlk: You can add it and read from viper config at cmd/sonrd/main.go

prdlk (changes requested):

proto/schema/v1/query.proto
joshLong145: > I think it's unnecessary to include both of these. Just the definition should be fine. Ill separate into a separate query method and proto for the `WhatIs`

proto/schema/v1/tx.proto
joshLong145: > This will still need the creator for now

yes ...this we will still need the Creator, it got caught in the fire.

x/schema/keeper/schema.go
joshLong145: i was under the impression we had seperate domains for writing and reading. ill change that

ntindle: We do, I thought ipfs_inter.add uses the same endpoint for both

x/schema/keeper/schema.go
joshLong145: the interface doesnt supper memory buffers from the looks up it as its parameters are a hash (cid) and file path. so i dont think we can support an in memory model albeit unfortunate.

x/schema/keeper/schema.go
joshLong145: updated to use two shells

x/schema/keeper/schema.go
joshLong145: updated to use `bufio` for in memory buffering of the file content a bit better, we still need to write to disk as the implantation of the underlying api is a bit weird with using tar ball unzips of the ipfs content.

prdlk (approved):
Looks good to me! 🚀

web/wallet/src/store/generated/package.json
joshLong145: the wallet directory was empty except for the package file it was a leftover artifact

ntindle (approved):

mcjcloud (approved):

x/schema/keeper/schema.go
```diff
@@ -1,14 +1,69 @@
...
+	ipfs_inter_write      = shell.NewShell(url_persistence_write)
+)
+
+func (k Keeper) LookUpContent(cid string, content interface{}) error {
+	time_stamp := string(rune(time.Now().Unix()))
+
+	out_path := filepath.Join(os.TempDir(), cid+time_stamp+".txt")
+	defer os.Remove(out_path)
+
+	err := ipfs_inter_read.Get(cid, out_path)
```
ntindle: You can do this through `ipfs.sonr.ws/ipfs/{CID}`

app.env.sample
```diff
@@ -9,4 +9,5 @@ FS_DIR=./
 PATH_PREFIX=/
 COSMOS_NODE_ADDRESS=http://localhost:26656  # Address of Cosmos backend
 LOCAL_CONFIG_DIR="~/.config/sonr" # a directory in which all local data is stored
-JWT_SECRET=BOBANDALICE
\ No newline at end of file
+JWT_SECRET=BOBANDALICE
+IPFS_API=:api.ipfs.sonr.ws:5001
```
ntindle: drop the 5001, its rewriting it internally

PR #403, opened 2022-06-27
Add password encrypted shard

mcjcloud:
Encrypt the recovery shard with a user-provided password before storing it in the vault.

## Changes

- `password.go` in `crypto` package.
- Update `CreateAccount` arguments. This will be changed further with introduction of protos.


internal/motor/motor.go
```diff
@@ -23,15 +23,27 @@ type MotorNode struct {
...
+	}
+
+	// TODO: encrypt dscShard with dsc (i.e. webauthn)
+	dscShard := string(deviceShard)
+
+	// TODO: ecnrypt pskShard with psk (must be generated)
+	pskShard := string(sharedShard)
+
+	// password protect the recovery shard
+	pwShard, err := crypto.EncryptWithPassword(password, []byte(recShard))
```
joshLong145, line 46: is it possible to generate the pwShard in a background process? this is going to see performance hits on lesser hardware with the current runtime targets

joshLong145:
looks good but a bit concerned with slow performance in our differernt client targets. we should consider if shard creation can be offboarded to a seperate thread.

internal/motor/motor.go
mcjcloud: I agree this needs optimization. Right now I'm wanting to make sure that these steps succeed before creating the WhoIs on chain. Created issue for this: #404.

PR #406, opened 2022-06-28
wrapping of wallet endpoints for motor in wasm compatibility

joshLong145:
-- addition of wasm wrapper for motor wallet endpoints

## Changes
* Item 1
* Item 2

## Checklist
* [ ] Unit tests
* [ ] Documentation

## References _(optional)_

Fixes 
Connects 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202529466067520) by [Unito](https://www.unito.io)

### DIFF
--- a/bind/motor-wasm/wasm.go
+++ b/bind/motor-wasm/wasm.go
@@ -1,18 +1,263 @@
+//go:build wasm
 // +build wasm
 
 package main
 
 import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"syscall/js"
+
 	"github.com/sonr-io/sonr/pkg/crypto"
+	"github.com/sonr-io/sonr/pkg/did"
+	"github.com/sonr-io/sonr/pkg/did/ssi"
 )
 
-func main() {
-	_, err := crypto.GenerateWallet()
-	check(err)
+var (
+	errWalletExists    = errors.New("mpc wallet already exists")
+	errWalletNotExists = errors.New("mpc wallet does not exist")
+)
+
+type motor struct {
+	//	node *host.SonrHost
+	wallet *crypto.MPCWallet
+	// host   host.SonrHost
 }
 
-func check(err error) {
-	if err != nil {
-		panic(err)
+var instance *motor
+
+// NewWallet creates a new mpc based wallet.
+func NewWallet() error {
+	if instance != nil {
+		return errWalletExists
 	}
+	w, err := crypto.GenerateWallet()
+	if err != nil {
+		return err
+	}
+	instance = &motor{
+		wallet: w,
+	}
+	return nil
+}
+
+// Address returns the address of the wallet.
+func Address() string {
+	if instance == nil {
+		return ""
+	}
+	addr, err := instance.wallet.Address()
+	if err != nil {
+		return ""
+	}
+	return addr
+}
+
+// LoadWallet unmarshals the given JSON into the wallet.
+func LoadWallet(buf []byte) error {
+	if instance != nil {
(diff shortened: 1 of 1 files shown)


bind/motor-wasm/wasm.go
```diff
@@ -1,18 +1,263 @@
...
+}
+
+// Address returns the address of the wallet.
+func Address() string {
+	if instance == nil {
+		return ""
+	}
+	addr, err := instance.wallet.Address()
+	if err != nil {
+		return ""
```
ntindle: Should we be returning the errors in some way instead of swallowing them

bind/motor-wasm/wasm.go
```diff
@@ -1,18 +1,263 @@
...
+}
+
+// MarshalWallet returns the JSON representation of the wallet.
+func MarshalWallet() []byte {
+	if instance == nil {
+		return nil
+	}
+	buf, err := instance.wallet.Marshal()
+	if err != nil {
+		return nil
```
ntindle: Why nil over an empty array?

bind/motor-wasm/wasm.go
```diff
@@ -1,18 +1,263 @@
...
+	}
+	return buf
+}
+
+// Sign returns the signature of the given message.
+func Sign(typeUrl string, msg []byte) ([]byte, error) {
+	if instance == nil {
+		return nil, errWalletNotExists
```
ntindle, line 127: This deviates from our paradigm above of not retuning two params

bind/motor-wasm/wasm.go
```diff
@@ -1,18 +1,263 @@
...
+func Verify(msg []byte, sig []byte) bool {
+	if instance == nil {
+		return false
+	}
+	return instance.wallet.Verify(msg, sig)
+}
+
+// Broadcast broadcasts rawTx to the specified address
+func Broadcast(addr string, tx []byte) error {
+	apiEndpoint := "v1-beta.sonr.ws:1317/cosmos/tx/v1beta/txs"
```
ntindle: This worries me but if it must be somewhere can we consolidate it to one central location?

ntindle (changes requested):

bind/motor-wasm/wasm.go
mcjcloud: +1, the binding should support errors. If wasm only supports single return values, we could return arrays that represent tuples.

bind/motor-wasm/wasm.go
mcjcloud: I think nil makes sense here actually. The idea won't be to operate on the contents of the array but just to deserialize it. So it's kind of an all-or-nothing sort of array.

mcjcloud (changes requested):
Consider making protos for requests/responses of these functions, since that's how target code will be interacting with go code.

bind/motor-wasm/wasm.go
joshLong145: yea we shouldnt be statically defining endpoints. this was taken from the `bind.go` implementations.

bind/motor-wasm/wasm.go
joshLong145: good call

bind/motor-wasm/wasm.go
joshLong145: the wrapper for the language target will filter to single value returns.

bind/motor-wasm/wasm.go
joshLong145: moved to a variable

prdlk (approved):

PR #452, opened 2022-06-30
Encrypt dsc and psk shards

mcjcloud:
Completes the CreateAccount function by adding the remaining shard encryption steps. Note that ECDSA cannot be used for encryption, only signing. This means AES keys will need to be provided 

**Changes**

* Add AES encryption methods
* Generate and use PSK, returning it to caller for storage and distribution
* Use provided password to encrypt recovery shard
* Use provided DSC AES key to encrypt DSC shard

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202535412858086) by [Unito](https://www.unito.io)


internal/motor/motor.go
```diff
@@ -7,10 +7,10 @@ import (
 	prt "go.buf.build/grpc/go/sonr-io/motor/registry/v1"
 )
 
-func CreateAccount(requestBytes []byte) (*registry.MotorNode, error) {
+func CreateAccount(requestBytes []byte) (prt.CreateAccountResponse, error) {
 	var request prt.CreateAccountRequest
 	if err := json.Unmarshal(requestBytes, &request); err != nil {
```
joshLong145, line 12: does this mean we should be encoding for json? perhaps create account remains agnostic of encoding standards and takes the object type directly.

internal/motor/proto/registry/v1/account.proto
```diff
@@ -6,11 +6,10 @@ option go_package = "github.com/sonr-io/sonr/motor/types/registry";
 // CreateAccount Request contains the three keys needed to create an account on Sonr
 message CreateAccountRequest {
   string password = 1;
-  bytes ecdsa_dsc_key = 2;
```
joshLong145, line 9: Comments on the purpose of each property would be helpful here

joshLong145 (approved):

prdlk (approved):

PR #472, opened 2022-07-14
Remove ctx from MotorNode

mcjcloud:
Remove ctx from MotorNode


prdlk:
LFG

PR #476, opened 2022-07-21
Feat/store refactoring

drgomesp:
## TODO

- [x] Refactor `ipfs.Protocol` (extract protocol and store interfaces)
- [x] Implement in-memory caching mechanism on `store.Store` 
- [x] Replace `keeper.Schema`'s IPFS shell calls with the refactored Store/IPFS protocol.
- [x] Refactor IPFSLite implementation


pkg/store/store.go
```diff
@@ -0,0 +1,13 @@
+package store
+
+import (
+	"github.com/ipfs/go-datastore"
+	"github.com/ipld/go-ipld-prime/storage"
+)
+
+// Store is a read-write storage with datastore.Batching support.
+type Store interface {
+	storage.ReadableStorage
```
joshLong145: Do we need a wrapper interface for the store? i feel using the provided data stores should be suffcient

pkg/protocol/ipfs_shell.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return i.shell.DagGet(ref, out)
+}
+
+func (i *IPFSShell) DagPut(ctx context.Context, data interface{}, inputCodec, storeCodec string) (string, error) {
+	panic("TODO")
+}
+
+func (i *IPFSShell) LookUpData(ctx context.Context, cid string, data interface{}) error {
```
drgomesp: We will remove this

pkg/protocol/ipfs_shell.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if err = json.Unmarshal(buf, &data); err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func (i *IPFSShell) PutData(ctx context.Context, data []byte) (string, error) {
```
drgomesp, line 77: Write small test case for checking if went through IPFS and failed at cache

internal/highway/x/protocol/ipfs_lite.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the file from IPFS
+	rsc, err := i.peer.GetFile(ctx, c)
+	if err != nil {
+		return nil, err
+	}
+
```
joshLong145, line 111: this can be removed

internal/highway/x/protocol/ipfs_lite.go
```diff
@@ -0,0 +1,203 @@
...
+type IPFSLite struct {
+	ctx       context.Context
+	node      host.SonrHost
+	dataStore *store.Memory
+	peer      *ipfslite.Peer
+	lock      sync.Mutex
+}
+
+func (i *IPFSLite) PinFile(ctx context.Context, cidstr string) error {
+	i.lock.Lock()
```
joshLong145: i still think the mutex belongs in the memory store layer. W should move `i` to memory store and lock in that layer

pkg/protocol/ipfs_shell.go
joshLong145: we should remove this method

joshLong145 (commented):
looking better, i still think we should be locking resources at a lower level than the interface implementations. If possible move the `i` mutex for the shell implementation to the store level  as we dont know when the cache will be hit if the defer unlocking of the resource will actually be at the correct time. and is also now up to the next implementor to make sure to lock resources if not at the store level.

pkg/protocol/ipfs_shell_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/stretchr/testify/mock"
+	"golang.org/x/sys/unix"
+
+	"github.com/sonr-io/sonr/pkg/protocol"
+	"github.com/sonr-io/sonr/x/schema/types"
+
+	"github.com/ipfs/go-datastore"
+)
+
+const IPFSShellUrl = "localhost:5001"
```
joshLong145, line 18: Can we move this to a configurable option?

joshLong145 (changes requested):
Looking good, can we revert the go mod and go sun changes? We should handle migrating versions in another pr. Also see comment left inline.

drgomesp:
@joshLong145 ready

ntindle:
Don't force push to our repo.

ntindle:
Can you migrate the code into a branch without force pushing? It won't be able to merge to dev @drgomesp

PR #480, opened 2022-07-25
Beta/schemas

joshLong145:
* addition of schema interfacing module in `pkg` allows for building of ipfs `nodes` for object constrcution. supports serialization as `dag-json` and `dag-cbor` can add additional encodings once we have support for encryption.
* below is the internal interface definition:

        type AppSchemaInternal interface {
        
        	/*
        		Builds a linkage of IPLD nodes from the provided schema definition
        		returns the `Node` and assigns it to the given id internally.
        	*/
        	BuildNodesFromDefinition(def *st.SchemaDefinition, object map[string]interface{}) (datamodel.Node, error)
        
        	/*
        		Returns an error if any of the keys within provided data dont match the given schema definition
        		useful for verifying
        	*/
        	VerifyObject(doc map[string]interface{}, def *st.SchemaDefinition) error
        
        	/*
        		Encodes a given IPLD Node as JSON
        	*/
        	EncodeDagJson(node datamodel.Node) ([]byte, error)
        
        	/*
        		Encodes a given IPLD Node as CBOR
        	*/
        	EncodeDagCbor(node datamodel.Node) ([]byte, error)
        
        	/*
        		Encodes a given IPLD Node as CBOR
        	*/
        	DecodeDagJson(buffer []byte) (datamodel.Node, error)
        
        	/*
        		Decodes a given IPLD Node as CBOR
        	*/
        	DecodeDagCbor(buffer []byte) (datamodel.Node, error)
        
        	/*
        		Get an IPLD object as a flat path of nodes.
        	*/
        	GetPath(node datamodel.Node) (datamodel.ListIterator, error)
        }## Changes
* Item 1
* Item 2

## Checklist
* [x] Unit tests
* [ ] Documentation

## References _(optional)_

Fixes 
Connects 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202665120657770) by [Unito](https://www.unito.io)


prdlk (approved):
Need module simulation back but LFG

pkg/schemas/events.go
```diff
@@ -0,0 +1,44 @@
...
+	ipfsApi "github.com/ipfs/go-ipfs-api"
+)
+
+type Event struct {
+	name     string
+	previous cid.Cid
+}
+
+func fetchEvent(shell ipfsApi.Shell, eventCid cid.Cid) (*Event, error) {
+	/*
```
mcjcloud: nit: commented out code?

pkg/schemas/schema_test.go
```diff
@@ -0,0 +1,175 @@
...
+		assert.NoError(t, err)
+
+		assert.NotNil(t, node)
+	})
+
+	t.Run("Should build Nodes from definition", func(t *testing.T) {
+		_, def := CreateMocks("snr12345", "did:snr:1234")
+		def.Fields = append(def.Fields, &st.SchemaKindDefinition{
+			Name:  "field-1",
+			Field: st.SchemaKind_INT,
```
mcjcloud, line 66: `Field` looks like it would more accurately be called `Type` or `Kind`.

pkg/schemas/schemas.go
```diff
@@ -0,0 +1,92 @@
...
+
+/*
+	Creates a relationship between a schema definition and the 'whatIs' definition
+*/
+type SchemaRelationShip struct {
+	definition *st.SchemaDefinition
+	did        did.DID
+}
+
+type appSchemaInternalImpl struct{}
```
mcjcloud: What is the purpose of this struct? It looks like all of its receiver methods aren't using the value at all.

This struct could store the schemas itself so that validating/building objects doesn't require the schema to be passed in.

mcjcloud (changes requested):
A few comments

pkg/schemas/schemas.go
joshLong145: I was thinking of keeping an internal map of schemas to did's which is what that relationship method is for. but im not in love with this module keeping internal state as that should be done by the `highway` or `motor`

pkg/schemas/events.go
joshLong145: removed for now

pkg/schemas/schemas.go
joshLong145: adding some internal query implementations for `whatIS` and query schema by cid

pkg/schemas/schemas.go
```diff
@@ -0,0 +1,105 @@
...
+	/*
+		Gets all `whatIs` objects for the account `whoIs` an error if the query fails
+	*/
+	GetSchemaByCid(cid string) ([]*st.SchemaKindDefinition, error)
+}
+
+/*
+	Creates a relationship between a schema definition and the 'whatIs' definition
+*/
+type SchemaRelationShip struct {
```
mcjcloud: nit: Relationship

pkg/schemas/schema_test.go
prdlk: @joshLong145 @mcjcloud maybe just keep it Kind like the enum type

mcjcloud (approved):

PR #485, opened 2022-08-01
Beta/Objects

joshLong145:
* Addition of internal object pkg for binding to higher level targets.
## Changes
* Item 1
* Item 2

## API Updates
following internal api is being proposed for objects. we can discuss adding an internal reference to `AppSchemaInternal` to allow for further helper methods to statically assert on an object with a provided schema. at this time the interface supports uploading and retrieving objects from `ipfs` storage.

```
type AppObjectInternal interface {
	/*
		Persists an object definition to the storage configured within the module.
	*/
	CreateObject(
		schema schemas.AppSchemaInternal,
		label string,
		object map[string]interface{}) (*ObjectUploadResult, error)

	/*
		Retrieves an object for the data store
	*/
	GetObject(cid string) (map[string]interface{}, error)
}
```

## Checklist
* [ ] Unit tests
* [ ] Documentation

## References _(optional)_

Fixes 
Connects 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202703507037753) by [Unito](https://www.unito.io)


x/registry/module_simulation.go
```diff
@@ -86,73 +86,74 @@ func (am AppModule) RegisterStoreDecoder(_ sdk.StoreDecoderRegistry) {}
...
-	simState.AppParams.GetOrGenerate(simState.Cdc, opWeightMsgDeactivateWhoIs, &weightMsgDeactivateWhoIs, nil,
-		func(_ *rand.Rand) {
-			weightMsgDeactivateWhoIs = defaultWeightMsgDeactivateWhoIs
-		},
-	)
-	operations = append(operations, simulation.NewWeightedOperation(
-		weightMsgDeactivateWhoIs,
-		registrysimulation.SimulateMsgDeactivateWhoIs(am.accountKeeper, am.bankKeeper, am.keeper),
-	))
+	/*
```
mcjcloud: Why is this all commented out?

pkg/object/object.go
```diff
@@ -0,0 +1,45 @@
...
+	Code       int32
+	Definition *ObjectDefinition
+	Message    string
+}
+
+type AppObjectInternalImpl struct {
+	shell *shell.Shell
+}
+
+func New(schemaImpl schemas.AppSchemaInternal, shell *shell.Shell) AppObjectInternal {
```
mcjcloud, line 35: You're not using the value of `schemaImpl` here. You could keep this in `AppObjectInternalImpl` and then remove the need to pass it into the create method.

mcjcloud (changes requested):
Two tiny comments.

This is so hype.

app.env.sample
```diff
@@ -10,4 +10,4 @@ PATH_PREFIX=/
 COSMOS_NODE_ADDRESS=http://localhost:26656  # Address of Cosmos backend
 LOCAL_CONFIG_DIR="~/.config/sonr" # a directory in which all local data is stored
 JWT_SECRET=BOBANDALICE
-IPFS_API=:api.ipfs.sonr.ws
\ No newline at end of file
+IPFS_API_READ =api.ipfs.sonr.ws
```
ntindle, line 13: Technically this is ipfs api write

ntindle (commented):
Where are the cmd commands

app.env.sample
joshLong145: indeed it is, ill update

pkg/object/object.go
joshLong145: that will work, but then each schema is 1-1 with an object.

x/registry/module_simulation.go
joshLong145: was for testing, i need to stop doing this.

pkg/object/object.go
mcjcloud: Good point.

mcjcloud (approved):

PR #486, opened 2022-08-02
Motor CreateSchema

mcjcloud:
Implement `CreateSchema` in motor

## Changes
* Add `CreateSchema` method

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202704187293220) by [Unito](https://www.unito.io)


internal/motor/create_schema.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return mt.CreateSchemaResponse{}, fmt.Errorf("broadcast tx: %s", err)
+	}
+
+	csresp := &st.MsgCreateSchemaResponse{}
+	if err := client.DecodeTxResponseData(resp.TxResponse.Data, csresp); err != nil {
+		return mt.CreateSchemaResponse{}, fmt.Errorf("decode MsgCreateSchemaResponse: %s", err)
+	}
+
+	return mt.CreateSchemaResponse{
```
joshLong145, line 48: we should return the full what is, as the schema module will require it.

joshLong145: We might want to create a map of did -> whatIs, and query for the `SchemaDefinition` from ipfs once the schema is created. this could be implemented in a `retrieveSchemaDefinition` method to be explicit.

internal/motor/create_schema.go
```diff
@@ -0,0 +1,67 @@
...
+		Did: csresp.WhatIs.Did,
+	}, nil
+}
+
+func convertFields(fields map[string]mt.CreateSchemaRequest_SchemaKind) ([]*st.SchemaKindDefinition, error) {
+	result := make([]*st.SchemaKindDefinition, len(fields))
+	var i int32
```
joshLong145, line 55: Do we need this variable?

joshLong145 (changes requested):
left a few comments. looking good.

internal/motor/create_schema.go
mcjcloud: Do you mean in the motor? I think a `GetSchema` method needs to exist in the chain.

As for keeping these cached in the motor, let's work out a design tomorrow (Wednesday) which allows us to cache schemas and update the cache.

internal/motor/create_schema.go
mcjcloud: It's being used on line 59 to update the `result` list. I believe populating the array this way is more efficient than using `append` when you know the length of the array ahead of time.

internal/motor/create_schema.go
mcjcloud: Added `WhatIs` to the response.

joshLong145 (approved):

PR #483, opened 2022-08-01
Motor moved from internal/ to pkg/

6missedcalls:
Motor moved from internal to pkg

## Changes
* Item 1
* Item 2

## Checklist
* [ ] Unit tests
* [ ] Documentation

## References _(optional)_

Fixes 
Connects 

<sub> Mention [stepsize] in a comment if you'd like to report some technical debt. See examples [here](https://app.stepsize.com/api/demo-pr-redirect).</sub>

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202703423429464) by [Unito](https://www.unito.io)

PR #487, opened 2022-08-02
Refactor motor

mcjcloud:
Refactor motor to make it consumable by other Go projects.

## Changes
* Move `internal/motor` to `pkg/motor`
* Convert `MotorNode` into an interface with default implementation

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202708961904895) by [Unito](https://www.unito.io)


bind/motor-mobile/bind.go
```diff
@@ -34,38 +35,67 @@ func Init(buf []byte) ([]byte, error) {
...
+	}
 }
 
 // Address returns the address of the wallet.
 func Address() string {
 	if instance == nil {
 		return ""
 	}
-	addr, err := instance.Wallet.Address()
+	addr, err := instance.GetWallet().Address()
```
joshLong145: we should split this into two calls and check for nil on the wallet return

bind/motor-mobile/bind.go
```diff
@@ -74,7 +104,7 @@ func Address() string {
 
 // Balance returns the balance of the wallet.
 func Balance() int {
-	return int(instance.Balance())
+	return int(instance.GetBalance())
```
joshLong145, line 111: should be nil checked for the cast and return. this could panic.

bind/motor-mobile/bind.go
```diff
@@ -94,7 +124,7 @@ func DidDoc() string {
 	if instance == nil {
 		return ""
 	}
-	buf, err := instance.DIDDocument.MarshalJSON()
+	buf, err := instance.GetDIDDocument().MarshalJSON()
```
joshLong145: same as above, should be nil checked

joshLong145 (changes requested):
looking good, but a few changes.

joshLong145:
@brokecollegekidwithaclothingobsession I think this PR will cover yours which migrates the motor to `pkg` @mcjcloud could you also add the changes. to the task file which removed to no longer defined task from running for `bind:ios` which is defined in #483

bind/motor-mobile/bind.go
mcjcloud: `GetBalance` can't return nil

bind/motor-mobile/bind.go
mcjcloud: done

bind/motor-mobile/bind.go
mcjcloud: done

joshLong145 (approved):

PR #490, opened 2022-08-03
add nil check

mcjcloud:
Fix DID() method

### DIFF
--- a/x/registry/keeper/msg_server_who_is.go
+++ b/x/registry/keeper/msg_server_who_is.go
@@ -95,7 +95,12 @@ func (k msgServer) UpdateWhoIs(goCtx context.Context, msg *types.MsgUpdateWhoIs)
 	for k, v := range msg.GetMetadata() {
 		val.Metadata[k] = v
 	}
-	controllerId, err := doc.GetController(val.DidDocument.DID())
+
+	d := val.DidDocument.DID()
+	if d == nil {
+		return nil, fmt.Errorf("error getting did from did document for creator '%s'", msg.Creator)
+	}
+	controllerId, err := doc.GetController(*d)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


x/registry/types/did.go
```diff
@@ -117,9 +121,9 @@ func stringifyVerificationRelationships(auth did.VerificationRelationships) ([]s
 	return res, nil
 }
 
-func (d *DIDDocument) DID() did.DID {
+func (d *DIDDocument) DID() *did.DID {
 	id, _ := did.ParseDID(d.Id)
```
SCKelemen, line 125: off-topic, but DIDDocument.Id should probably be DIDDocument.ID to follow not only Go convention, but to be internally consistent as well.
```suggestion
	id, _ := did.ParseDID(d.ID)
```

SCKelemen (approved):

x/registry/keeper/msg_server_who_is.go
```diff
@@ -95,7 +95,12 @@ func (k msgServer) UpdateWhoIs(goCtx context.Context, msg *types.MsgUpdateWhoIs)
 	for k, v := range msg.GetMetadata() {
 		val.Metadata[k] = v
 	}
-	controllerId, err := doc.GetController(val.DidDocument.DID())
+
+	d := val.DidDocument.DID()
```
SCKelemen, line 99: Why id it DidDocument here but DIDDocument further down?

6missedcalls (approved):
LGTM

x/registry/types/did.go
mcjcloud: We have a lot of inconsistencies and code cleanup to do after we get our initial launch done. Can you file a bug? Same for the below comment.

PR #495, opened 2022-08-04
Ref motor schema object

mcjcloud:
Refactor schema/object interface for motor

## Changes

* Move internal tools into internal/
* Rename interfaces/implementations
* Return struct pointers instead of interfaces from constructors


joshLong145 (approved):

ntindle (approved):

prdlk (approved):
LGTM 🚀

PR #496, opened 2022-08-04
Add external facing object builder

mcjcloud:
Adds an external facing object builder that users of the package can use to interact with the underlying schema/object interface.

## Changes

- Add `pkg/motor/x/object`

### New Features

Add struct `ObjectBuilder`

```go
type ObjectBuilder struct {}

func (ob *ObjectBuilder) SetLabel(string)
func (ob *ObjectBuilder) Set(string, interface{}) error
func (ob *ObjectBuilder) Remove(string) bool
func (ob *ObjectBuilder) Build() (Object, error)
```

The object type represents an object ready to be uploaded to the network. It can be passed into `objectImpl.CreateObject(...)`.

An object builder accepts a `Schema` and `object.ObjectClient` interface during construction, allowing users to provide their own implementations.


ntindle (approved):

pkg/motor/x/object/schema_interface.go
```diff
@@ -0,0 +1,20 @@
...
+	GetValue() int32
+}
+
+type SchemaField interface {
+	GetName() string
+	GetKind() SchemaKind
+	TryValue(interface{}) bool
+}
+
+type Schema interface {
```
joshLong145, line 16: this will need to be updated my `beta/schema-links`

joshLong145 (approved):
Looks good, interfaces will need updating with schema linking but this looks good

PR #499, opened 2022-08-04
Schema Link Support (Sub Schemas)

joshLong145:
## Changes
* Addition of link support for complex type support in `Schemas`. supports pre processing schema for link of schema type.
* Loads schemas from cid link and caches in `storage` implementation.

## Checklist
* [x] Unit tests
* [ ] Documentation

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202724946484339) by [Unito](https://www.unito.io)


internal/schemas/links.go
```diff
@@ -1,29 +1,51 @@
...
+				return errSchemaFieldsInvalid
+			}
+			links = append(links, f.Link)
+		}
+	}
+
+	for len(links) > 0 {
+		key := links[len(links)-1]
+		links = links[:len(links)-1]
+		buf, err := as.store.Get(context.Background(), key)
```
mcjcloud: `err` isn't being used

proto/common/link.proto
```diff
@@ -5,7 +5,9 @@ package sonrio.sonr.common;
 option go_package = "github.com/sonr-io/sonr/x/common/types";
 
 enum LinkKind {
```
mcjcloud, line 7: What is the purpose of bucket and object here?

x/schema/keeper/grpc_query_schema.go
```diff
@@ -27,13 +27,14 @@ func (k Keeper) Schema(goCtx context.Context, req *st.QuerySchemaRequest) (*st.Q
...
+	var schemaJson *st.SchemaDefinition
+	err := k.LookUpContent(what_is.Schema.Cid, schemaJson)
+	fields := make([]*st.SchemaKindDefinition, len(schemaJson.Fields))
+	for _, v := range schemaJson.Fields {
 		fields = append(fields, &st.SchemaKindDefinition{
-			Name:  k,
-			Field: v.(st.SchemaKind),
+			Name: v.Name,
+			// here we need to downcast the
+			Field: interface{}(v.Field).(st.SchemaKind),
```
mcjcloud: Is this cast necessary? `v.Field` is already `SchemaKind`

x/schema/keeper/schema.go
```diff
@@ -17,11 +16,11 @@ import (
 )
 
 var (
-	url        = "https://api.ipfs.sonr.ws"
```
mcjcloud, line 20: Did you mean to commit this?

x/schema/keeper/schema.go
```diff
@@ -40,7 +39,7 @@ func (k Keeper) LookUpContent(cid string, content interface{}) error {
 		return err
 	}
 
-	if err = json.Unmarshal(buf, &content); err != nil {
+	if err = content.Unmarshal(buf); err != nil {
```
mcjcloud, line 48: If `content` is nil, unmarshal may panic.

mcjcloud (changes requested):

proto/common/link.proto
joshLong145: these will be used for objects, and buckets to denote the link "kind" such that we can delimitate the context of the `cid`

x/schema/keeper/schema.go
joshLong145: good point

x/schema/keeper/schema.go
joshLong145: yes, for now its safer to statically define the route rather than going through the cli env variables. this might fail when deployed as i found the env variables where not loading properly.

mcjcloud (approved):
lgtm

internal/schemas/schema_test.go
```diff
@@ -1,14 +1,134 @@
...
 )
 
+func CreateMockHeirachyThreeLevel(creator string, did string) (st.WhatIs, st.SchemaDefinition) {
+
+	mockWhatIs := st.WhatIs{
+		Did: did,
+		Schema: &st.SchemaReference{
+			Did:   did,
+			Label: "testing schema",
+			Cid:   "asdasd12312",
```
ntindle, line 22: This isn't a real CID

ntindle (approved):

PR #500, opened 2022-08-05
Create object builder for motor

mcjcloud:
Creates an object builder for external motor users and caches schemas that have been created/fetched

## Changes
* Add `NewObjectBuilder` in `pkg/motor` which acts as default implementation
* Move `ObjectClient` interface into `pkg`
* Store schemas and WhatIs locally in motor instance
* Add `UploadObject` to Motor interface (without implementation for now)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202725549863154) by [Unito](https://www.unito.io)


pkg/motor/resource.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (r *motorResources) GetSchema(did string) (*st.WhatIs, *st.SchemaDefinition, bool) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if _, ok := r.whatIsStore[did]; !ok {
+		return nil, nil, false
+	}
+	if def, ok := r.schemaStore[r.whatIsStore[did].Schema.Cid]; ok {
```
joshLong145: nit: Although we check above for the key to be defined, i thnk checks on the Schema object being non nil is a good safe guard

internal/object/interface.go
```diff
@@ -18,11 +33,4 @@ type ObjectClient interface {
 		Retrieves an object for the data store
 	*/
 	GetObject(cid string) (map[string]interface{}, error)
-
```
joshLong145, line 21: was assert removed from the interface?

pkg/motor/resource.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/// StoreWhatIs fetches
+func (r *motorResources) StoreWhatIs(whatIs *st.WhatIs) (*st.SchemaDefinition, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.whatIsStore[whatIs.Did] = whatIs
+
+	if schema, ok := r.schemaStore[whatIs.Schema.Cid]; ok {
```
joshLong145, line 38: We should check on the schema object being non nil before accessing `Cid`

joshLong145 (changes requested):
few comments, looking good.

internal/object/interface.go
mcjcloud: It didn't look like it was being used, is it necessary?

prdlk (approved):

pkg/motor/object_builder_test.go
```diff
@@ -0,0 +1,77 @@
...
+prt.CreateSchemaRequest{
+	Label: "TestUser",
+	Fields: map[string]prt.CreateSchemaRequest_SchemaKind{
+		"email":     prt.CreateSchemaRequest_SCHEMA_KIND_STRING,
+		"firstName": prt.CreateSchemaRequest_SCHEMA_KIND_STRING,
+		"age":       prt.CreateSchemaRequest_SCHEMA_KIND_INT,
+	},
+}
+*/
+const SCHEMA_DID string = "did:snr:f2604d3b-4a30-4223-ba63-070a075b7db8"
```
ntindle: Can I run this in a pipeline?

mcjcloud: Yes so long as the chain state doesn't get removed. I added TODO for making a better test suite

pkg/motor/object_builder.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func (mtr *motorNodeImpl) NewObjectBuilder(did string) (*object.ObjectBuilder, error) {
+	whatIs, schema, found := mtr.resources.GetSchema(did)
+	if !found {
+		return nil, fmt.Errorf("could not find WhatIs with did '%s'", did)
+	}
+
+	schemaImpl := schemas.New(schema.Fields, whatIs)
+	objCli := objectcli.New(schemaImpl, shell.NewShell("https://api.ipfs.sonr.ws"))
```
joshLong145: nit: uri should not be inlined

pkg/motor/resource.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// StoreWhatIs fetches
+func (r *motorResources) StoreWhatIs(whatIs *st.WhatIs, cb SchemaCallback) {
+	callback := func(s *st.SchemaDefinition, e error) {
+		r.mu.Unlock()
+		cb(s, e)
+	}
+
+	go func() {
```
joshLong145: im a bit confused on why this is running in a go routine. is this to load the schema in a background context? it might be better to load in the same context and invoke the callback in the same execution context.

pkg/motor/resource.go
```diff
@@ -0,0 +1,111 @@
...
+		r.mu.Lock()
+
+		r.whatIsStore[whatIs.Did] = whatIs
+
+		if whatIs.Schema == nil {
+			callback(nil, fmt.Errorf("WhatIs '%s' has no schema", whatIs.Did))
```
joshLong145: nit: errors should be constants in the package

joshLong145 (changes requested):

pkg/motor/resource.go
mcjcloud: The idea is that when you create/fetch schemas they will probably not be needed right away. If you think it's better to just wait for them I can get rid of the callback.

pkg/motor/resource.go
mcjcloud: Yes I agree. One thing with constants is that we can't easily include details like the DID. I noticed we have a lot of inconsistent errors in the repo so I created #505 to deal with this.

joshLong145: sounds good to me and i agree. resolving

pkg/motor/resource.go
joshLong145: i think we should introduce it later if we need this kind of implementation. im in favor of the callback approach but think execution should stay synchronous

internal/object/interface.go
joshLong145: it was intended for the motor to wrap in order to provide a static assert endpoint for objects. it will build the object and assert on it and return an error if it doesnt comply with schemas. I should add some documentation so that it was more obvious.

mcjcloud: If it's a helper method then the interface doesn't necessarily need it because other implementations may do things differently.

pkg/motor/resource.go
mcjcloud: sgtm, done

pkg/motor/object_builder.go
mcjcloud: done

joshLong145:
Looks good

joshLong145 (approved):

PR #501, opened 2022-08-05
Introduce ADR-3 (Buckets) & Clean Pkg

prdlk:
This introduces the Categorical Encrypted Storage implementation ("Buckets") into the Sonr Blockchain. The purpose of this  module is to provide a mechanism to be able to store developer app data, and user app data into secure encrypted IPFS instances.

## Changes

* Create `x/bucket`
* Introduce `BucketItem`
* Introduce `WhereIs`

### Deprecations

* Removed `common.LinkKind`, any 3rd party protobufs need to be put in a `third_party` directory in the root-level directory of the project. This Enum was only utilized by the schema package and I have moved its proto type into `proto/schema`
* Removed `proto/common` - see above

### Enhancements

* Restructured `pkg/crypto` the pkg folder contained many packages that would be associated with "cryptography". I have moved the existing files in `pkg/crypto` into a directory called `pkg/crypto/mpc`
* Renamed `MPCWallet` to `Wallet`. Now that the package is called `mpc` we can simply remove the prefix
* Moved `did`, `jwt`, `jwx` into `pkg/crypto` all cryptographic packages are now subdirectories of crypto

## Next Steps

* [ ] Create TX Msg's for Bucket Module
* [ ] Implement Bucket Keeper
* [ ] Implement Bucket Query Methods: `QueryByLabel`, `QueryByDid`, `QueryByCreator`, `QueryUserBucketByApp`
* [ ] Implement Bucket Client
* [ ] Build IPFS Service for Bucket Generation

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202730550404233) by [Unito](https://www.unito.io)


proto/bucket/where_is.proto
```diff
@@ -0,0 +1,57 @@
...
+  // Role of the creator of the new bucket.
+  WhereIsRole role = 5;
+
+  // IsActive flag of the new bucket.
+  bool is_active = 6;
+
+  // Content of the new bucket map of DIDs to CIDs.
+  map<string, string> content = 7;
+
+  // JWKs of the new bucket.
```
ntindle, line 28: Why need

x/bucket/genesis_test.go
```diff
@@ -14,6 +14,15 @@ func TestGenesis(t *testing.T) {
 	genesisState := types.GenesisState{
 		Params: types.DefaultParams(),
 
+		WhereIsList: []types.WhereIs{
+			{
+				Did: "did:sonr:1",
+			},
+			{
+				Did: "did:sonr:2",
```
ntindle, line 22: Lets make these real dids

proto/bucket/genesis.proto
```diff
@@ -3,12 +3,15 @@ package sonrio.sonr.bucket;
...
 import "bucket/params.proto";
+import "bucket/where_is.proto";
 // this line is used by starport scaffolding # genesis/proto/import
 
 option go_package = "github.com/sonr-io/sonr/x/bucket/types";
 
 // GenesisState defines the bucket module's genesis state.
 message GenesisState {
   Params params = 1 [(gogoproto.nullable) = false];
+  repeated WhereIs whereIsList = 2 [(gogoproto.nullable) = false];
```
mcjcloud: This should be underscore case `where_is_list` (same for any other protos)

proto/bucket/query.proto
```diff
@@ -27,4 +38,21 @@ message QueryParamsResponse {
   Params params = 1 [(gogoproto.nullable) = false];
 }
 
+message QueryGetWhereIsRequest {
+	string did = 1;
+}
+
+message QueryGetWhereIsResponse {
+	WhereIs WhereIs = 1 [(gogoproto.nullable) = false];
```
mcjcloud: Should be `where_is`

bind/motor-wasm/wasm.go
```diff
@@ -21,7 +21,7 @@ var (
 
 type motor struct {
 	//	node *host.SonrHost
-	wallet *crypto.MPCWallet
+	wallet *mpc.Wallet
```
ntindle, line 24: Does this remove the ability to have traditional wallets?

proto/bucket/where_is.proto
mcjcloud: This is for ACL

proto/bucket/where_is.proto
```diff
@@ -0,0 +1,72 @@
...
+  BucketVisibility visibility = 4;
+
+  // Role of the creator of the new bucket.
+  BucketRole role = 5;
+
+  // IsActive flag of the new bucket.
+  bool is_active = 6;
+
+  // JWKs of the new bucket.
+  map<string, bytes> jwks = 7;
```
mcjcloud: This should be a 2D map. I'd suggest a message for the Jwk map:
```
message AclJwks {
  map<string, bytes> encrypted_jwks = 1;
}

map<string, AclJwks> content_acl = 1;

mcjcloud (changes requested):

proto/bucket/where_is.proto
joshLong145: agree with @mcjcloud comment

joshLong145 (changes requested):
agreed with brayden on creating a message wrapper for jwk's

mcjcloud (approved):
lgtm

ntindle (approved):

PR #508, opened 2022-08-10
Feature/query what is by creator

joshLong145:
Addition of query what is by controllerr, returns all `WhatIs` definitions for the requested account address.
Addition of a check for `CreateSchema` on the `Definition`'s `Fields` for a length of 0. which returns a 400 status code.

## Changes
* Addition of whatIs by controller query in the blockchain `Schema` module.

## API Updates
* addition of query api endpoint
### Deprecations _(required)_
Schema which where defined as empty cannot be recreated as it will resulting in a `400` status code for non complete request. (missing at least one field). 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202780624906297) by [Unito](https://www.unito.io)


go.sum
```diff
@@ -2744,8 +2744,6 @@ github.com/zeebo/pcg v1.0.0 h1:dt+dx+HvX8g7Un32rY9XWoYnd0NmKmrIzpHF7qiTDj0=
 github.com/zeebo/pcg v1.0.0/go.mod h1:09F0S9iiKrwn9rlI5yjLkmrug154/YRW6KnnXVDM/l4=
 github.com/zondax/hid v0.9.0 h1:eiT3P6vNxAEVxXMw66eZUAAnU2zD33JBkfG/EnfAKl8=
 github.com/zondax/hid v0.9.0/go.mod h1:l5wttcP0jwtdLjqjMMWFVEE7d1zO0jvSPA9OPZxWpEM=
-go.buf.build/grpc/go/sonr-io/motor v1.4.28 h1:qK20CxFR6CZU4ZRnTgUEf7NIMFj40msOkf5NujVcxRg=
-go.buf.build/grpc/go/sonr-io/motor v1.4.28/go.mod h1:6SsVxJlKbyf3gkEZ7pNsOXOG9kOab5P17JuhvvOeUZM=
```
ntindle: Why do these keep gettin g added/removed

ntindle (approved):

proto/schema/v1/query.proto
```diff
@@ -75,4 +87,9 @@ message QueryWhatIsResponse {
   WhatIs whatIs = 1;
 }
 
+message QueryWhatIsCreatorResponse {
+  // Schema Repersentation
+  repeated WhatIs whatIs = 1;
```
mcjcloud: The field needs to be underscore cased `what_is = 1;`

x/schema/keeper/msg_server_schema.go
```diff
@@ -17,6 +17,15 @@ func (k msgServer) CreateSchema(goCtx context.Context, msg *types.MsgCreateSchem
 	if err != nil {
 		return nil, err
 	}
+	if len(msg.Definition.Fields) < 1 {
+		k.Logger(ctx).Info("Create schema request empty, must have defined fields. aborting operation")
+		return &types.MsgCreateSchemaResponse{
+			Code:    400,
```
mcjcloud: nit: Use http status codes `http.StatusInvalidRequest`

mcjcloud (changes requested):

x/schema/keeper/msg_server_schema.go
joshLong145: good call

go.sum
joshLong145: when protos are rebuilt it ups the minor version counter

x/schema/keeper/msg_server_schema.go
```diff
@@ -86,7 +83,7 @@ func (k msgServer) CreateSchema(goCtx context.Context, msg *types.MsgCreateSchem
 	k.SetWhatIs(ctx, whatIs)
 
 	resp := types.MsgCreateSchemaResponse{
-		Code:    200,
+		Code:    http.StatusAccepted,
```
mcjcloud, line 86: This is a `202` not `200`. Not that it's actually wrong...

mcjcloud (approved):
lgtm

PR #511, opened 2022-08-11
Add object builder bindings to mobile

mcjcloud:
Adds motor functions in motor-mobile for interacting with the object builder

Note: Because of Cgo, the interface will be more explicit than when using Go directly. Each target language can use these functions to create an abstraction.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202786913154580) by [Unito](https://www.unito.io)


bind/motor-mobile/object_builder.go
```diff
@@ -0,0 +1,116 @@
...
+#include <stdlib.h>
+*/
+import "C"
+
+func NewObjectBuilder(n, d *C.char) error {
+	if instance == nil {
+		return errWalletNotExists
+	}
+
+	name := C.GoString(n)
```
joshLong145, line 26: Can this panic or throw and error? we probably should check n and d for nil

bind/motor-mobile/object_builder.go
```diff
@@ -0,0 +1,116 @@
...
+	objectBuilders[name] = builder
+	return nil
+}
+
+func SetObjectLabel(n, l *C.char) error {
+	if instance == nil {
+		return errWalletNotExists
+	}
+
+	name := C.GoString(n)
```
joshLong145, line 52: same as a above on the nil check

joshLong145 (commented):
Left a few comments inline on some examples. think we need more nil checks on the pointers since all this data will have to be passed by reference its important to check for nil refs.

bind/motor-mobile/object_builder.go
mcjcloud: Good catch, done.

pkg/motor/test_keys/psksnr1nu3qhdvz84kpfg604ys9pc8n4cvmgxyn84vyhw
```diff
@@ -0,0 +1,2 @@
+C�
```
ntindle: Add to .gitignore

ntindle (approved):

joshLong145 (approved):

pkg/motor/test_keys/psksnr1nu3qhdvz84kpfg604ys9pc8n4cvmgxyn84vyhw
mcjcloud: done

PR #516, opened 2022-08-15
Feat/bucket chain impl

### LINKED ISSUE #471: Define bucket categories for app-specific, mailbox, and vault
## To-Do

* [x] Update [ADR-3](https://docs.sonr.io/posts/architecture/adr-003) to contain categories of buckets
* [x] Update `proto/bucket` to match [ADR-3](https://docs.sonr.io/posts/architecture/adr-003) CRUD methods
* [x] Push updated types to [Buf.build](buf.build/sonr-io/sonr)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202588794949851) by [Unito](https://www.unito.io)

joshLong145:
Initial implementations of `bucket` blockchain module

## API Updates
* CreateWhereIs
* UpdateWhereIs
* DeleteWhereIs
* queryWhereIs
* queryWhereIsByCreator

## Checklist
* [x] Unit tests
* [x] Documentation

## References _(optional)_

Fixes 
Connects 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202528620406598/1202802081607095) by [Unito](https://www.unito.io)


proto/bucket/query.proto
```diff
@@ -39,20 +45,29 @@ message QueryParamsResponse {
...
-	WhereIs where_is = 1 [(gogoproto.nullable) = false];
+  WhereIs where_is = 1 [(gogoproto.nullable) = false];
+}
+
+message QueryGetWhereIsByCreatorRequest {
+  string creator = 1;
+}
+
+message QueryGetWhereIsByCreatorResponse {
+  repeated WhereIs where_is = 1 [(gogoproto.nullable) = false];
```
ntindle, line 62: May want pagination?

proto/bucket/tx.proto
```diff
@@ -1,20 +1,27 @@
...
+  // this line is used by starport scaffolding # proto/tx/rpc
 }
 
 message MsgCreateWhereIs {
   string creator = 1;
+  string label = 2;
+  BucketVisibility visibility = 3;
+  BucketRole role = 4;
+  repeated BucketItem content = 5;
+  map<string, AclJwks> content_acl = 6;
```
ntindle, line 24: CC: @SCKelemen 

do we want access control for this release?

prdlk (approved):

pkg/ipns/builder.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func NewBuilder() *IPNSURIBuilder {
+	return &IPNSURIBuilder{
+		prefix: "ipfs",
+		delim:  "/",
+	}
+}
+
+func (iub *IPNSURIBuilder) WithCid(cid string) {
```
mcjcloud: nit: This should be `SetCid` since it doesn't return an `IPNSURIBuilder`. "with" makes me think it's a struct builder.

pkg/ipns/builder.go
```diff
@@ -0,0 +1,40 @@
...
+	iub.cid = cid
+}
+
+func (iub *IPNSURIBuilder) BuildString() string {
```
mcjcloud: If we name this `String` it will implement the Stringer interface which would be a plus.

pkg/ipns/builder.go
```diff
@@ -0,0 +1,40 @@
...
+	return fmt.Sprintf("%s%s%s%s", iub.delim, iub.prefix, iub.delim, iub.cid)
+}
+
+func (iub *IPNSURIBuilder) BuildService() did.Service {
+	url := ssi.MustParseURI(fmt.Sprintf("%s%s%s%s", iub.delim, iub.prefix, iub.delim, iub.cid))
```
mcjcloud: nit: looks like you could just use `iub.BuildString()` here.

pkg/ipns/ipns.go
```diff
@@ -0,0 +1,33 @@
...
+	id, err := shell.ID()
+	if err != nil {
+		return "", err
+	}
+	resp, err := shell.PublishWithDetails(rec.Builder.cid, id.ID, time.Hour, rec.Ttl, true)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Print(resp.Name)
```
mcjcloud: nit: Are these print statements used for something?

pkg/ipns/record.go
```diff
@@ -0,0 +1,47 @@
+package ipns
+
+import (
+	"time"
+
+	"github.com/ipfs/go-ipns"
+	pb "github.com/ipfs/go-ipns/pb"
+	crypto "github.com/libp2p/go-libp2p-core/crypto"
```
mcjcloud: Is there a reason to be using libp2p crypto vs built in crypto?

thirdparty/proto/v1/ipns/ipns_entry.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package sonrio.sonr.thirdparty.ipns;
+
+option go_package = "github.com/sonr-io/sonr/thirdparty/proto/v1/ipns";
```
mcjcloud: `thirdparty` should be `third_party` in the project structure if we're trying to stay conventional.

thirdparty/proto/v1/ipns/ipns_entry.proto
```diff
@@ -0,0 +1,23 @@
...
+
+message IpnsEntry {
+  enum ValidityType {
+    // setting an EOL says "this record is valid until..."
+    EOL = 0;
+  }
+  required bytes value = 1;
```
mcjcloud: Everything is optional by default in proto3 and it's considered bad practice to have required fields.

https://stackoverflow.com/a/31814967/4092920

x/bucket/keeper/grpc_query_where_is_by_creator.go
```diff
@@ -0,0 +1,27 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func (k Keeper) WhereIsByCreator(c context.Context, req *types.QueryGetWhereIsByCreatorRequest) (*types.QueryGetWhereIsByCreatorResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	if req == nil {
```
mcjcloud: duplicated nil check

x/bucket/keeper/msg_server_where_is.go
```diff
@@ -11,9 +12,41 @@ import (
 
 func (k msgServer) CreateWhereIs(goCtx context.Context, msg *types.MsgCreateWhereIs) (*types.MsgCreateWhereIsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	err := msg.ValidateBasic()
+	k.Logger(ctx).Info("basic request validation finished")
```
mcjcloud: nit: Printing this before the error check may be misleading

x/bucket/keeper/msg_server_where_is.go
```diff
@@ -11,9 +12,41 @@ import (
...
+	accts := msg.GetSigners()
+	if len(accts) < 1 {
+		k.Logger(ctx).Error("Error while querying account: not found")
+		return nil, sdkerrors.ErrNotFound
+	}
+
+	for _, c := range msg.Content {
+		if c.Type == types.ResourceIdentifier_CID {
+			if msg.ContentAcl[c.Uri] == nil {
+				k.Logger(ctx).Info("Content does not have an associated ACL %s", c.Uri)
```
mcjcloud, line 31: You're logging when there's no ACL for a content, but not doing anything else. Is this loop necessary? Maybe an empty map should be created for any missing ACLs.

mcjcloud (changes requested):

pkg/ipns/builder.go
joshLong145: thats a good point.

pkg/ipns/record.go
joshLong145: go-ipns requires keys from `go-libp2p-core`

x/bucket/keeper/msg_server_where_is.go
joshLong145: i was thinking about that but im wondering if the chain should be doing such operations as we originally talked about the motor being the only encryption point for content.

pkg/ipns/record.go
joshLong145: was able to convert keys over, think they wrapped this to add an equality operation to the interface -_-

ntindle (approved):

mcjcloud (approved):
lgtm

PR #520, opened 2022-08-19
Feat/bucket motor impl

joshLong145:
- addition of motor binding for internal bucket implementations
- Addition of `Create` and `Update` methods for `WhereIs` Instances

## API Updates
Addition of the following methods for motor interface
```
	CreateBucket(context.Context, mt.CreateBucketRequest) (bucket.Bucket, error)
	QueryWhereIs(context.Context, mt.QueryWhereIsRequest) (mt.QueryWhereIsResponse, error)

	UpdateBucketItems(context context.Context, did string, items []*bt.BucketItem) (bucket.Bucket, error)
	GetBucket(context context.Context, did string) (bucket.Bucket, error)
	GetBuckets(creator string) ([]bucket.Bucket, error)
```
### New Features *(required)*
- Creation of `WhereIs`
- Update of `WhereIs`

## Checklist

- [ ] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 


buf.work.yaml
```diff
@@ -1,4 +1,3 @@
 version: v1
 directories:
   - proto
-  - pkg/motor/proto
```
ntindle, line 4: ?

pkg/motor/motor.go
```diff
@@ -73,7 +52,7 @@ func EmptyMotor(id string) *motorNodeImpl {
 
 func initMotor(mtr *motorNodeImpl, options ...mpc.WalletOption) (err error) {
 	// Create Client instance
-	mtr.Cosmos = client.NewClient(client.ConnEndpointType_BETA)
+	mtr.Cosmos = client.NewClient(client.ConnEndpointType_LOCAL)
```
ntindle: ? Maybe let's not hardcode this?

buf.work.yaml
joshLong145: that directory got moved out and running `buf build` would error

pkg/motor/motor.go
joshLong145: .. this should had been reverted. you are right ill create an issue

internal/bucket/access.go
```diff
@@ -25,3 +25,25 @@ func (b *bucketImpl) GetBucketItems() []*bt.BucketItem {
 func (b *bucketImpl) ContentExists(id string) bool {
 	return b.contentCache[id] == nil
 }
+
+func (b *bucketImpl) GetContent() []*BucketContent {
+	var content []*BucketContent = make([]*BucketContent, len(b.contentCache))
+	for _, v := range b.contentCache {
+		if v.ContentType == bt.ResourceIdentifier_CID {
+			content = append(content, v)
```
mcjcloud, line 33: using `append` with a slice that has a preconfigured size will cause the slice to be longer than necessary. So in this `content` will have size `len(b.contentCache) * 2`

This applies below as well

pkg/motor/bucket_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil, err
+		}
+
+		mtr.Resources.whereIsStore[did] = wiReq.WhereIs
+	}
+	wi := mtr.Resources.whereIsStore[did]
+	s := mtr.Resources.shell
+	bq := mtr.Resources.bucketQueryClient
+
+	b := bucket.New(addr, wi, s, bq)
```
mcjcloud: nit: making local variables for each parameter here makes this less readable. Consider either making the variable names more obvious or just putting the parameters inline

pkg/motor/bucket_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/sonr-io/sonr/internal/bucket"
+	"github.com/sonr-io/sonr/pkg/motor/types"
+	bt "github.com/sonr-io/sonr/x/bucket/types"
+)
+
+func (mtr *motorNodeImpl) NewBucketResolver(context context.Context, did string) (bucket.Bucket, error) {
```
mcjcloud: In what circumstance would you call this vs `GetBucket`?

pkg/motor/create_bucket.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if request.Creator == "" {
+		return nil, errors.New("Invalid Address")
+	}
+
+	if request.Label == "" {
+		return nil, errors.New("Label nust be defined")
+	}
+
+	createWhereIsRequest := bt.NewMsgCreateWhereIs(request.Creator)
```
mcjcloud: This constructor should take all of the needed values

pkg/motor/create_bucket.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, fmt.Errorf("broadcast tx: %s", err)
+	}
+
+	cbresp := &bt.MsgCreateWhereIsResponse{}
+	if err := client.DecodeTxResponseData(resp.TxResponse.Data, cbresp); err != nil {
+		return nil, fmt.Errorf("decode MsgCreateWhereIsResponse: %s", err)
+	}
+
+	if cbresp.Status != http.StatusAccepted {
+		return nil, errors.New("Non success status from Create bucket Request")
```
mcjcloud: Include the actual status code here for debugging

pkg/motor/resource.go
```diff
@@ -58,6 +71,20 @@ func (r *motorResources) StoreWhatIs(whatIs *st.WhatIs) (*st.SchemaDefinition, e
 	return definition, nil
 }
 
+func (r *motorResources) StoreWhereIs(whereis *bt.WhereIs) bool {
+	_, ok := r.whatIsStore[whereis.Did]
```
mcjcloud, line 78: Did you mean to check `ok` before storing the whereis?

pkg/motor/ssi.go
```diff
@@ -0,0 +1,21 @@
+package motor
+
+import (
+	"errors"
+)
+
+func (mtr *motorNodeImpl) AddBucketServiceEndpoint(id string) error {
```
mcjcloud, line 11: We'll need some way of committing these changes. Not sure if `updateWhoIs` needs to be done here or someplace else.

pkg/motor/update_bucket.go
```diff
@@ -0,0 +1,95 @@
...
+func (mtr *motorNodeImpl) UpdateBucket(request mt.UpdateBucketRequest) (bucket.Bucket, error) {
+	if request.Creator == "" {
+		return nil, errors.New("Invalid Address")
+	}
+
+	if request.Label == "" {
+		return nil, errors.New("Label nust be defined")
+	}
+
+	createWhereIsRequest := bt.NewMsgUpdateWhereIs(request.Creator, request.Did)
```
mcjcloud: Same as above, I would update the constructor here to take all required fields

mcjcloud (changes requested):

pkg/motor/resource.go
joshLong145: my thinking was the cache should be updated regardless on call and the boolean is returned to indicate if the cache was already present

pkg/motor/bucket_resolver.go
joshLong145: this is used internally as a helper method. It could be moved to the bucket internal package instead.

pkg/motor/interface.go
```diff
@@ -0,0 +1,44 @@
...
+
+type MotorNode interface {
+	GetDeviceID() string
+
+	GetAddress() string
+	GetBalance() int64
+
+	GetClient() *client.Client
+	GetWallet() *mpc.Wallet
+	GetPubKey() *secp256k1.PubKey
```
SCKelemen, line 25: Are we sure we want to expose a specific Key type here? It seems like it will be difficult to change  in the future.

SCKelemen (approved):

pkg/motor/bucket_resolver.go
mcjcloud: I think it should then or be made private

pkg/motor/bucket_resolver.go
```diff
@@ -72,32 +63,20 @@ func (mtr *motorNodeImpl) GetBucket(context context.Context, did string) (bucket
 	return mtr.Resources.bucketStore[did], nil
 }
 
+/*
+	Takes the whereIs store and checks for a matching bucket in the cache, if its not present it will create it and get its sub buckets
+	Does not query for new buckets, only respects what is currently present in the store
```
mcjcloud, line 40: This should probably query instead of creating. If someone is calling "Get" they won't expect new resources to be created

pkg/motor/resource.go
```diff
@@ -101,3 +104,43 @@ func (r *motorResources) GetSchema(did string) (*st.WhatIs, *st.SchemaDefinition
 
 	return nil, nil, false
 }
+
+func (r *motorResources) GetWhereIs(ctx context.Context, did string, address string) error {
```
mcjcloud: Consider checking in cache for the value before fetching

pkg/motor/update_bucket.go
```diff
@@ -8,26 +8,24 @@ import (
...
-	"github.com/sonr-io/sonr/pkg/motor/types"
-	mt "github.com/sonr-io/sonr/pkg/motor/types"
 	"github.com/sonr-io/sonr/pkg/tx"
 	bt "github.com/sonr-io/sonr/x/bucket/types"
 )
 
-func (mtr *motorNodeImpl) UpdateBucket(request mt.UpdateBucketRequest) (bucket.Bucket, error) {
-	if request.Creator == "" {
-		return nil, errors.New("Invalid Address")
+func (mtr *motorNodeImpl) UpdateBucket(label string, did string, role bt.BucketRole, visibility bt.BucketVisibility, content []*bt.BucketItem) (bucket.Bucket, error) {
```
mcjcloud: Why are we pulling everything out of the request model only to put it back? This will require changes in binding layers too

pkg/motor/update_bucket.go
```diff
@@ -8,26 +8,24 @@ import (
...
+	if label == "" {
+		return nil, errors.New("label nust be defined")
 	}
 
-	createWhereIsRequest := bt.NewMsgUpdateWhereIs(request.Creator, request.Did)
-	createWhereIsRequest.Label = request.Label
-	createWhereIsRequest.Role = request.Role
-	createWhereIsRequest.Visibility = request.Visibility
-	createWhereIsRequest.Content = request.Content
+	createWhereIsRequest := bt.NewMsgUpdateWhereIs(mtr.Address, did)
```
mcjcloud: nit: should be called `updateWhereIsRequest`

mcjcloud (changes requested):

pkg/motor/resource.go
mcjcloud: Oh I see, resolving.

ntindle:
Adding a query filter for schema for the buckets query

I would see this being used as such

`/bucket/getContent?schema=did:snr:0922390u&creator=did:snr:a92u3jlfa&bucket=did:snr:asoidkfj32n09`
```
{
 "schema": "did:snr:0922390u",
 "creator": "did:snr:a92u3jlfa",
 "bucket": "did:snr:asoidkfj32n09"
}
```


---
List buckets by the creator is also a request. We may want to work on the naming for these endpoints as they aren't super user accessible

---
List content for all buckets by creator

---

This is a speedway request

pkg/motor/resource.go
joshLong145: `GetWhereIs` is responsible for updating the cache

pkg/motor/update_bucket.go
joshLong145: I can revert to using an object in the parameters for the binding layers

joshLong145:
> Adding a query filter for schema for the buckets query
> 
> I would see this being used as such
> 
> `/bucket/getContent?schema=did:snr:0922390u&creator=did:snr:a92u3jlfa&bucket=did:snr:asoidkfj32n09`
> 
> ```
> {
>  "schema": "did:snr:0922390u",
>  "creator": "did:snr:a92u3jlfa",
>  "bucket": "did:snr:asoidkfj32n09"
> }
> ```
> 
> List buckets by the creator is also a request. We may want to work on the naming for these endpoints as they aren't super user accessible
> 
> List content for all buckets by creator
> 
> This is a speedway request

This is a valid request. Will Create another PR for this feature. Will be implemented motor side with a model update to add `SchemaURI` which will contain the reference to a `SchemaDefinition`

pkg/motor/interface.go
joshLong145: these are valid points on the current `motor` interface. but out of scope of the current pr

ntindle: @SCKelemen open an issue for this. May want to fix this soonish. Do you have a better idea of how to handle it?

mcjcloud (approved):
lgtm

PR #522, opened 2022-08-19
Integrate motor for mobile

mcjcloud:
This PR contains changes needed in order to integrate motor into iOS and Android.

## Changes

- Create proto gen configs for Java and Swift
- Currently disable `SonrHost` in motor as it causes android to not work
- Update ObjectBuilder request types


prdlk:
@mcjcloud Check out how im doing it for [iOS/Android in flutter](https://github.com/sonr-io/motor-flutter/tree/upgrade/motor-v0.3.1/scripts), i made this about 30 mins ago.

ntindle (commented):
Why were all the nil checking and error messages removed?

buf.gen.yaml
```diff
@@ -1,7 +1,13 @@
 version: v1
 plugins:
-  - name: js
```
prdlk: This file should be in your android/ios/flutter type definition folders. Here are some examples:
- [motor-flutter buf.build generation script](https://github.com/sonr-io/motor-flutter/blob/main/scripts/proto.sh)
- [motor-flutter buf.gen.yaml file](https://github.com/sonr-io/motor-flutter/blob/main/lib/buf.gen.yaml)
- [MotorKit buf.gen.yaml file](https://github.com/sonr-io/MotorKit/blob/main/Sources/MotorKit/buf.gen.yaml)

pkg/motor/interface.go
```diff
@@ -25,7 +24,7 @@ type MotorNode interface {
 	GetPubKey() *secp256k1.PubKey
 	GetDID() did.DID
 	GetDIDDocument() did.Document
-	GetHost() host.SonrHost
+	// GetHost() host.SonrHost
```
prdlk: This doesnt need to be commented out because of my PR to isolate the libp2p connection process. On android you would simply not call `Connect()`

prnk28: On android you would simply not call `Connect()`

pkg/motor/motor.go
```diff
@@ -96,10 +95,10 @@ func initMotor(mtr *motorNodeImpl, options ...mpc.WalletOption) (err error) {
 	mtr.DID = *baseDid
 
 	// It creates a new host.
-	mtr.SonrHost, err = host.NewDefaultHost(context.Background(), config.DefaultConfig(config.Role_MOTOR))
-	if err != nil {
-		return err
-	}
+	// mtr.SonrHost, err = host.NewDefaultHost(context.Background(), config.DefaultConfig(config.Role_MOTOR))
```
prdlk: Update this branch from dev to see these changes in the new method

bind/motor-mobile/object_builder.go
```diff
@@ -1,34 +1,20 @@
...
 	_ "golang.org/x/mobile/bind"
 )
 
-/*
-#include <stdlib.h>
-*/
-import "C"
-
-func NewObjectBuilder(n, d *C.char) error {
+func NewObjectBuilder(name, schemaDid string) error {
```
prdlk, line 9: Is it possible to bind an `ObjectBuilder` interface whose methods are exported to the target language? This could probably shoulder off a lot of code duplication. [See this article](https://medium.com/@matryer/tutorial-calling-go-code-from-swift-on-ios-and-vice-versa-with-gomobile-7925620c17a4)

prdlk (changes requested):

bind/motor-mobile/object_builder.go
```diff
@@ -1,34 +1,20 @@
...
 	if instance == nil {
```
SCKelemen, line 10: Where does instance come from?

bind/motor-mobile/object_builder.go
```diff
@@ -1,34 +1,20 @@
...
 	}
-	if n == nil {
-		return errors.New("name cannot be nil")
-	}
-	if d == nil {
-		return errors.New("schema did cannot be nil")
-	}
 
-	name := C.GoString(n)
 	if _, ok := objectBuilders[name]; ok {
```
SCKelemen, line 14: I don't see objectBuilders coming from anywhere either? Are these automatically sourced from bind or somewhere else in the package?

proto/schema/v1/schema.proto
```diff
@@ -4,6 +4,9 @@ package sonrio.sonr.schema;
 
 option go_package = "github.com/sonr-io/sonr/x/schema/types";
 
+option java_package = "io.sonr.types.schema";
```
SCKelemen, line 7: Should this be `"io.sonr.schemas.types"`

bind/motor-mobile/object_builder.go
mcjcloud: Yes, for this and the above, look in `motor-mobile/bind.go`

```

var (
	instance       mtr.MotorNode
	objectBuilders map[string]*object.ObjectBuilder
	callback       MotorCallback
)
```

proto/schema/v1/schema.proto
mcjcloud: The project structure in Java doesn't separate out the modules like that, it has:
* io.sonr.motoroid
and that's it

So rather than creating a package for each domain that contains only a `types` package, it makes more sense to create one `types` package which contains all the domains

bind/motor-mobile/object_builder.go
mcjcloud: That makes sense. I will look into this in a subsequent PR. Filed #535

prdlk (approved):

mcjcloud:
> Why were all the nil checking and error messages removed?

@ntindle Because I replaced the C-style types with Go types, which cannot be nil.

ntindle (approved):

PR #524, opened 2022-08-22
Feat/motor logger

joshLong145:
* Addition of `golog` in `pkg/motor` for structured logging

## Changes

* refactoring of `fmt` print statements to `golog` structured logging format.

## API Updates

* Addition of `loglevel` paramater to `EmptyMotor`

## Checklist

* [ ] Unit tests
* [ ] Documentation

## References *(optional)*

Fixes
Connects

### DIFF
--- a/pkg/motor/motor.go
+++ b/pkg/motor/motor.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/kataras/golog"
 	"github.com/mr-tron/base58"
 	"github.com/sonr-io/multi-party-sig/pkg/party"
 	"github.com/sonr-io/sonr/pkg/client"
@@ -51,7 +52,7 @@ type motorNodeImpl struct {
 	DID         did.DID
 	DIDDocument did.Document
 	SonrHost    host.SonrHost
-
+	Logger      *golog.Logger
 	// Sharding
 	deviceShard   []byte
 	sharedShard   []byte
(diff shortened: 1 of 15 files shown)


pkg/motor/motor.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/kataras/golog"
```
mcjcloud, line 9: I believe it was the built in "log" package which automatically supports various targets. This package looks nice, but have we confirmed it works on all targets?

pkg/motor/object_builder.go
```diff
@@ -10,8 +10,10 @@ import (
 )
 
 func (mtr *motorNodeImpl) NewObjectBuilder(did string) (*object.ObjectBuilder, error) {
+	mtr.Logger.Info("Building object with schema: %s", did)
```
mcjcloud, line 13: This method doesn't actually build an object but just creates the object builder.

mcjcloud (changes requested):
I think logging every time an error is returned may be overkill since the error messages get composed together and printed out already.

pkg/motor/motor.go
```diff
@@ -65,15 +66,22 @@ type motorNodeImpl struct {
 	Resources *motorResources
 }
 
-func EmptyMotor(id string) *motorNodeImpl {
+func EmptyMotor(id string, logLevel string) *motorNodeImpl {
```
prdlk, line 69: After this is updated to `dev` you could add log level as a property in the `InitializeRequest` [protobuf file](https://github.com/sonr-io/sonr/blob/dev/proto/motor/v1/request.proto). This would keep it inline with existing method.

prdlk (changes requested):

pkg/motor/motor.go
joshLong145: i was between this and `zap` but i think youre right. we will loose structured logging however.

ntindle:
Can we update this from dev if we still plan to do it?

joshLong145:
closing to due to another branch continuing this functionality. will link to that pr here.

joshLong145:
work has been continued in PR: #583

PR #525, opened 2022-08-22
Feat/query bucket by schema id

joshLong145:
Adds the ability to query by Schema identifier for content within a `bucket` which matches the given Schema.
## Changes
- Addition of `QueryBucketbySchema` within `pkg/motor`
- Support for json marshal/unmarshal on `BucketContent`
- Migration of `BucketContent` to protobuf definition. 
## API Updates
`QueryBucketBySchema` `pkg/motor``

### New Features *(required)*

- Ability to query for content implementing a given schema.
## Checklist

- [ ] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 


pkg/motor/query_bucket.go
```diff
@@ -34,3 +36,51 @@ func (mtr *motorNodeImpl) QueryWhereIsByCreator(ctx context.Context) (mt.QueryWh
 		WhereIs: resp,
 	}, nil
 }
+
+func (mtr *motorNodeImpl) QueryBucketBySchema(ctx context.Context, req mt.QueryBucketContentBySchemaRequest) (mt.QueryBucketContentBySchemaResponse, error) {
+	if req.Creator != mtr.Address {
+		return mt.QueryBucketContentBySchemaResponse{}, fmt.Errorf("creator address does not match session creator: %s", mtr.Address)
+	}
```
ntindle: Do we need this?

ntindle (approved):

pkg/motor/query_bucket.go
mcjcloud: +1 You should be able to query schemas you didn't create.

This feature in general will be improved when we add namespacing

internal/bucket/access.go
```diff
@@ -47,3 +47,19 @@ func (b *bucketImpl) GetBuckets() []*BucketContent {
 
 	return content
 }
+
+func (b *bucketImpl) ResolveContentBySchema(did string) ([]*BucketContent, error) {
+	var matchedContent []*BucketContent = make([]*BucketContent, 0)
+	for _, c := range b.contentCache {
+		if c.ContentType == bt.ResourceIdentifier_CID {
+			bi, ok := c.Item.(*bt.BucketItem)
+			if ok {
```
mcjcloud: nit: You can reduce nesting by doing `if ok && bi.SchemaDefinition.Did == did {`

You could even combine the above line (if you want) to 
```
if bi, ok := c.Item.(*bt.BucketItem); ok && bi.SchemaDefinition.Did == did {
}
```

internal/bucket/bucket.go
```diff
@@ -24,10 +25,19 @@ var (
...
-	Item        interface{}
-	Id          string
-	ContentType bt.ResourceIdentifier
+	Item        interface{}           `json:"item"`
+	Id          string                `json:"id"`
+	ContentType bt.ResourceIdentifier `json:"contentType"`
 }
+
+func (bc *BucketContent) MarshalJson() ([]byte, error) {
+	return json.Marshal(bc)
```
mcjcloud: To implement `Marshaler` the function should be `MarshalJSON`. With the json tags above however, it should be unnecessary to implement. Instead you can do `json.Marshal(...)` in the calling code.

internal/bucket/bucket.go
```diff
@@ -24,10 +25,19 @@ var (
...
+}
+
+func (bc *BucketContent) UnmarshalJson(data []byte, ref *map[string]interface{}) error {
```
mcjcloud: Same comment as above.

mcjcloud (changes requested):

internal/bucket/access.go
joshLong145: i went for the above to improve readability of the conditions but what you are proposing isnt that bad. ill update

internal/bucket/bucket.go
joshLong145: right, i added this for ease when serializing for protobuf and other requests that might need the data in such a format. it was brought in to make it easier on the consumer. ill make the changes to 'marshalJSON' and 'unmarshalJSON'

pkg/motor/query_bucket.go
joshLong145:
```
if b.WhereIs.Visibility == types.BucketVisibility_PRIVATE && b.WhereIs.Creator != mtr.Address {
  return mt.QueryBucketContentBySchemaResponse{}, fmt.Errorf("creator address does not match session creator: %s", mtr.Address)
}`
we still need to filter private schemas
```

pkg/motor/interface.go
```diff
@@ -37,6 +37,7 @@ type MotorNode interface {
 	CreateBucket(context.Context, mt.CreateBucketRequest) (bucket.Bucket, error)
 	QueryWhereIs(ctx context.Context, did string) (mt.QueryWhereIsResponse, error)
 	QueryWhereIsByCreator(ctx context.Context) (mt.QueryWhereIsByCreatorResponse, error)
+	QueryBucketBySchema(context.Context, mt.QueryBucketContentBySchemaRequest) (mt.QueryBucketContentBySchemaResponse, error)
```
prdlk: This can be consolidated into the generic query request found in the motor protobuf file, after updating the branch

pkg/motor/resource.go
```diff
@@ -125,6 +125,18 @@ func (r *motorResources) GetWhereIs(ctx context.Context, did string, address str
 
 	r.whereIsStore[res.WhereIs.Did] = res.WhereIs
 
+	b := bucket.New(address, r.whereIsStore[res.WhereIs.Did], r.shell, r.bucketQueryClient)
```
prdlk: Bucket Querying interface should be short-lived grpc connections

pkg/motor/resource.go
```diff
@@ -142,6 +154,18 @@ func (r *motorResources) GetWhereIsByCreator(ctx context.Context, address string
 	for _, wi := range res.WhereIs {
 		r.whereIsStore[wi.Did] = &wi
 		ptrArr = append(ptrArr, &wi)
+
+		b := bucket.New(address, r.whereIsStore[wi.Did], r.shell, r.bucketQueryClient)
```
prdlk: Maybe we possibly try simplifying the creation of a bucket with just pkg/client?

proto/motor/v1/response.proto
```diff
@@ -4,6 +4,7 @@ syntax = "proto3";
 package sonrio.motor.api.v1;
 
 import "bucket/where_is.proto";
+import "google/protobuf/any.proto";
```
prdlk: Why the unused import?

proto/motor/v1/request.proto
```diff
@@ -112,3 +107,17 @@ message UpdateBucketRequest {
   sonrio.sonr.bucket.BucketRole role = 5;
   repeated sonrio.sonr.bucket.BucketItem content = 6;
 }
+
+message QueryWhereIsRequest {
+  string creator = 1;
+  string did = 2;
+}
+
+message QueryBucketContentBySchemaRequest {
```
prdlk: @ntindle @mcjcloud @brokecollegekidwithaclothingobsession @joshLong145  We need to figure out a better naming convention for query methods collectively lol

internal/bucket/bucket.go
```diff
@@ -24,9 +24,9 @@ var (
 	CID -> reference to content (map[string]interface{})
 */
 type BucketContent struct {
```
prdlk, line 28: If this is added as a protobuf type within `proto/common` we would have easier Json serialization, and extensibility with the definition across other modules within motor.

prdlk (changes requested):

pkg/motor/resource.go
joshLong145: Thats probably a better idea, i can migrate some of this logic there with the merger of #528

proto/motor/v1/response.proto
joshLong145: should be removed from a previous implementation

internal/bucket/bucket.go
joshLong145: I think youre right that this should be moved to a proto def, I think the same can be said about some of the `Object` structs as well.

6missedcalls (approved):
LGTM

internal/bucket/access.go
```diff
@@ -37,13 +38,40 @@ func (b *bucketImpl) GetContent() []*BucketContent {
...
 	}
 
 	return content
 }
+
+func (b *bucketImpl) ResolveContentBySchema(did string) ([]*mt.BucketContent, error) {
+	var matchedContent []*mt.BucketContent = make([]*mt.BucketContent, 0)
+	for _, c := range b.whereIs.Content {
+		if c.Type == bt.ResourceIdentifier_CID {
+			if c.SchemaDefinition.Did == did {
```
mcjcloud: If you're only checking resources with CIDs, won't none of the DID fields be present?

proto/bucket/where_is.proto
```diff
@@ -54,6 +56,10 @@ message BucketItem {
 
   // Type of the resource
   ResourceIdentifier type = 4;
+
+  // References the schema and item within the bucket is associated with. Bucket items do not need to use the same schema reference
+  // if the bucket is "generic" // optional
+  sonrio.sonr.schema.SchemaReference schema_definition = 5;
```
mcjcloud: nit: name this `schema_reference` since "SchemaDefinition" is its own thing.

proto/motor/v1/bucket.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
```
mcjcloud: I think this file would fit better in the bucket types rather than motor types.

mcjcloud (changes requested):

internal/bucket/access.go
joshLong145: the identifier check is to make sure that the content is not a bucket, then the check on the given schema did is to check if the schema being queried from outside the scope is referenced by content within the specified bucket. `SchemaReference` is now just a `DID` as the entire reference was not needed

proto/bucket/where_is.proto
joshLong145: refactored to just be the `schema_did`

proto/motor/v1/bucket.proto
joshLong145: migrated

mcjcloud (approved):

prdlk (approved):

PR #542, opened 2022-08-26
Feat motor android

mcjcloud:
Fixes to motor needed for querySchema in android.


scripts/bind.sh
```diff
@@ -22,7 +22,7 @@ while getopts "iawm" opt; do
       rm -rf ${BUILDDIR}/io.sonr.motor-sources.jar
       cd ${BUILDDIR}
       tar -czvf motor-${VERSION}-android.tar.gz io.sonr.motor.aar
-      rm -rf ${ANDROID_ARTIFACT}
```
ntindle, line 25: ?

pkg/client/query.go
```diff
@@ -10,41 +10,16 @@ import (
 	"google.golang.org/grpc"
 )
 
-// func (c *Client) QueryAccount(address string) (*at.BaseAccount, error) {
```
ntindle, line 13: Does speedway use this?

pkg/motor/query.go
```diff
@@ -6,45 +6,10 @@ import (
 
 	mt "github.com/sonr-io/sonr/third_party/types/motor"
 	bt "github.com/sonr-io/sonr/x/bucket/types"
+	st "github.com/sonr-io/sonr/x/schema/types"
 )
 
-func (mtr *motorNodeImpl) QueryBucket(req mt.QueryWhereIsRequest) (*mt.QueryWhereIsResponse, error) {
```
ntindle, line 11: Will this break speedway?

pkg/client/query.go
mcjcloud: It was commented out so I don't think so.

scripts/bind.sh
mcjcloud: This was deleting the .aar file which I need.

pkg/motor/query.go
mcjcloud: Not until speedway upgrades version, it will be a simple rename

ntindle (approved):

prdlk (approved):

PR #544, opened 2022-08-30
Integrate object builder into motoroid

mcjcloud:
Changes needed to integrate ObjectBuilder into Motoroid

## Changes

- Add build and upload methods to binding layer
- Replace object builder Go types with proto types
- Add QueryObject method to motor


prdlk (approved):

pkg/motor/query.go
```diff
@@ -105,3 +106,11 @@ func (mtr *motorNodeImpl) QueryWhereIsByCreator(req mt.QueryWhereIsByCreatorRequ
 		WhereIs: ptrArr,
 	}, nil
 }
+
+func (mtr *motorNodeImpl) QueryObject(cid string) (map[string]interface{}, error) {
+	sh := shell.NewShell(mtr.Cosmos.GetIPFSApiAddress())
```
joshLong145: there is a instance of shell already scoped within the `Motor`.

bind/motor-mobile/object_builder.go
```diff
@@ -129,3 +130,54 @@ func RemoveObjectField(name, fieldName string) error {
...
+}
+
+func UploadObject(name string) ([]byte, error) {
+	if instance == nil {
+		return nil, errWalletNotExists
+	}
+
+	builder, ok := objectBuilders[name]
+	if !ok {
+		return nil, fmt.Errorf("no object builder with name '%s'", name)
```
joshLong145, line 160: nit: perhaps we should move this a global var for object errors.

mcjcloud: I'm not sure what you mean. A global object for storing errors that occur in motor?

pkg/motor/query.go
mcjcloud: I did not notice an instance in Motor but I added a reference to `motorNodeImpl`.

ntindle (approved):
Work with Ian to make sure this doesn't break speedway

bind/motor-mobile/object_builder.go
joshLong145: I mena more make it a global var in the bind layer. but we dont have this setup really yet. Ill check to see if there is an issue. resolving this comment.

joshLong145 (approved):

PR #546, opened 2022-08-30
Query schema by did

mcjcloud:
Query schemas by DID alone.

## Changes

- Add WhatIsByDid method to chain
- Return empty list from SchemaByCreator instead of nil


x/schema/keeper/schema.go
```diff
@@ -119,6 +119,26 @@ func (k Keeper) GetWhatIsFromCreator(ctx sdk.Context, creator string) (val []*ty
...
+
+	for ; iterator.Valid(); iterator.Next() {
+		instance := &types.WhatIs{}
+		if err := k.cdc.Unmarshal(iterator.Value(), instance); err != nil {
+			return nil, false
+		}
+		if instance.Did == did {
+			return instance, true
+		}
+	}
```
joshLong145: I could be wrong but i dont think you need to iterate over the store, i think you can get the 'WhatIs' from the keeper of the 'DID' as that is they key used for content within the schema keeper.

mcjcloud: Good catch, fixed.

joshLong145 (approved):

ntindle:
Does this change things for speedway

ntindle (approved):

joshLong145 (approved):

mcjcloud:
No changes for speedway @ntindle, unless you want to add a similar query method

PR #583, opened 2022-09-13
Ref/motor logging

joshLong145:
addition of `motor/x/logger` for logging within `pkg/motor`

## Changes

* refactored all `fmt` print statements to use the logger now within `motorNodeImpl`
* added `pkg/motor/x/logger` for all internal logging within the motor package
* logger has levels and ability to format strings.

## Checklist

* [ ] Unit tests
* [ ] Documentation

## References *(optional)*

Fixes
Connects


pkg/motor/x/logger/logger.go
```diff
@@ -0,0 +1,64 @@
...
+		level:    conevertLogLevel(level),
+		category: category,
+		logger:   log.New(os.Stdout, category, 0),
+	}
+}
+
+func (l *Logger) log(level LogLevel, category string, message string, args ...interface{}) {
+	if l.level <= level {
+		ts := time.Now().Format("2006-01-02 15:04:05")
+		prefix := fmt.Sprintf("%s %s %s : ", conevertLogLevelCategory(level), ts, l.category)
```
mcjcloud: You accept `category` as a parameter but don't use it.

pkg/motor/x/logger/logger.go
```diff
@@ -0,0 +1,64 @@
...
+	CAT_FATAL = "[FATAL]"
+)
+
+type Logger struct {
+	logger   *log.Logger
+	level    LogLevel
+	category string
+}
+
+func New(level string, category string) *Logger {
```
mcjcloud: It looks like the `category` isn't actually used anywhere. Does it need to be part of the struct?

pkg/motor/x/logger/utils.go
```diff
@@ -0,0 +1,35 @@
+package logger
+
+func conevertLogLevelCategory(level LogLevel) string {
+	switch level {
+	case LEVEL_INFO:
+		return "[INFO]"
```
mcjcloud: nit: Use the constants here

pkg/motor/x/logger/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	case LEVEL_ERROR:
+		return "[ERROR]"
+	case LEVEL_FATAL:
+		return "[FATAL]"
+	}
+
+	return ""
+}
+
+func conevertLogLevel(level string) LogLevel {
```
mcjcloud: Where does the string version of these levels come from?

pkg/motor/motor.go
```diff
@@ -67,10 +77,12 @@ func EmptyMotor(r *mt.InitializeRequest, cb common.MotorCallback) (*motorNodeImp
 		supportDir:         r.GetSupportDir(),
 		tempDir:            r.GetTempDir(),
 		callback:           cb,
+		logLevel:           r.LogLevel,
 	}, nil
 }
 
 func initMotor(mtr *motorNodeImpl, options ...mpc.WalletOption) (err error) {
+	mtr.log = logger.New(mtr.logLevel, "motor")
```
mcjcloud, line 84: I'm confused as to what the second argument here is for.

mcjcloud (changes requested):
I think this package would be useful outside of motor, like directly in `pkg/`

pkg/motor/x/logger/utils.go
joshLong145: it comes from motor as to allow for strings in the request for initialize

joshLong145:
> I think this package would be useful outside of motor, like directly in `pkg/`

has been moved to 'pkg'

prdlk (approved):

prdlk:
> I think this package would be useful outside of motor, like directly in `pkg/`

Maybe internal?

proto/motor/v1/request.proto
```diff
@@ -168,3 +170,19 @@ message SeachBucketContentBySchemaRequest {
 }
 
 message StatRequest {}
+
+message UploadDocumentRequest {
```
mcjcloud, line 174: Why is this in this PR?

joshLong145: think it was from a merger i did and committed by mistake, good catch. removing this

mcjcloud (approved):

PR #591, opened 2022-09-14
Introduce/schema document

prdlk:
## Changes

* Item 1
* Item 2

## Checklist

* [ ] Unit tests
* [ ] Documentation

## References *(optional)*

Fixes
Connects


x/schema/types/document.go
```diff
@@ -0,0 +1,122 @@
...
+			},
+		}
+	case string:
+		if strings.Contains(v, "did:") && name != "@did" {
+			return &SchemaDocumentValue{
+				Name:  name,
+				Field: SchemaKind_LINK,
+				LinkValue: &LinkValue{
+					Value: v,
+					Link:  LinkKind_UNKNOWN,
```
joshLong145, line 100: I think the link kind would be `LinkKind_SCHEMA` in this context

x/schema/types/document.go
```diff
@@ -0,0 +1,122 @@
...
+			Field: SchemaKind_STRING,
+			StringValue: &StringValue{
+				Value: v,
+			},
+		}
+	case []*SchemaDocumentValue:
+		return &SchemaDocumentValue{
+			Name:  name,
+			Field: SchemaKind_LIST,
+			ArrayValue: &ArrayValue{
```
joshLong145, line 115: for a type of `List` I think we need Value and Item type to facilitate template generation to be contextless.

proto/schema/v1/document.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message ArrayValue {
+    repeated SchemaDocumentValue value = 1;
+}
+
+
+message SchemaDocumentValue {
+    string name = 1;
+    // Type of a single schema property
+    SchemaKind field = 2;
```
joshLong145, line 60: i think we need an optional field for array item types something like `ItemKind` which would be defined something like the following: 
message ItemKindDefinition {
 SchemaKind type = 1
 ItemKindDefinition item = 2;
}
this definition is recursive to accommodate multidimensional arrays

pkg/motor/login.go
```diff
@@ -47,13 +47,13 @@ func (mtr *motorNodeImpl) LoginWithKeys(request mt.LoginWithKeysRequest) (mt.Log
 	mtr.Address = request.Did
 
 	// fetch vault shards
-	mtr.callback.OnMotorEvent("Fetching shards from vault", false)
+	// TODO: Breaking (bind.ios) mtr.callback.OnMotorEvent("Fetching shards from vault", false)
```
joshLong145, line 50: could there be a check on the callback instance for the `OnMotorEvent` being nil instead of commenting out?

joshLong145 (changes requested):

bind/motor-mobile/bind.go
```diff
@@ -19,7 +19,6 @@ type MotorCallback interface {
 var (
 	objectBuilders map[string]*object.ObjectBuilder
 	instance       mtr.MotorNode
-	callback       ct.MotorCallback
```
mcjcloud, line 22: Why is this being removed?

pkg/motor/login.go
mcjcloud: +1

pkg/motor/object_builder.go
```diff
@@ -19,3 +21,48 @@ func (mtr *motorNodeImpl) NewObjectBuilder(did string) (*object.ObjectBuilder, e
...
+}
+
+func (mtr *motorNodeImpl) UploadDocument(req mt.UploadDocumentRequest) (*mt.UploadDocumentResponse, error) {
+	builder, err := mtr.NewObjectBuilder(req.GetDefinition().GetDid())
+	if err != nil {
+		return nil, err
+	}
+
+	builder.SetLabel(req.GetLabel())
+	builder.Set("@did", req.GetDefinition().GetDid())
```
mcjcloud, line 47: `@did` is likely to not be a field in whichever schema is being used. This should cause an error during upload or building.

proto/schema/v1/document.proto
```diff
@@ -0,0 +1,70 @@
...
+
+    // The value of the field
```
mcjcloud, line 62: Have you considered protobuf's `oneof` field for this?

third_party/types/common/callback.go
```diff
@@ -7,7 +7,9 @@ type MotorCallback interface {
 	OnMotorEvent(msg string, isDone bool)
 }
 
-type defaultCallback struct{}
+type defaultCallback struct {
+	MotorCallback
```
mcjcloud: Remove this line as it silences errors whenever `defaultCallback` does not implement `MotorCallback`.

mcjcloud (dismissed):

pkg/motor/object_builder.go
joshLong145: interesting, we could add a clause that '@' prefixed keys are not checked in the schema verification

proto/schema/v1/document.proto
joshLong145: cosmos doesnt support it currently

prdlk: Since we structured protos in same place as cosmos opposed to third party we cannot use oneof as a feature

pkg/motor/object_builder.go
prdlk: yes

prdlk: we need a way to relate back to the schema

bind/motor-mobile/bind.go
prdlk: It doesnt need to be there since its being passed to the Motor Node

prdlk:
This needs to get merged

proto/schema/v1/document.proto
joshLong145: will be addressed in follow up implementation around schemas

joshLong145 (approved):

ntindle (approved):

PR #611, opened 2022-09-19
added configs for localnet

lakshya-sky:
## Adds
- `test-config.yml`
- `docker/Dockerfile.localnet`

This files will be used by scripts in https://github.com/sonr-io/local-dev-tools for bootstrapping localnet for development purposes.

### DIFF
--- a/test-config.yaml
+++ b/test-config.yaml
@@ -1,0 +1,21 @@
+accounts:
+  - mnemonic: "satisfy adjust timber high purchase tuition stool faith fine install that you unaware feed domain license impose boss human eager hat rent enjoy dawn"
+    name: validator
+    coins: ["20000snr", "100000000stake"]
+    # address: snr1phaxpevm5wecex2jyaqty2a4v02qj7qmqe4lvm
+  - mnemonic: "notice oak worry limit wrap speak medal online prefer cluster roof addict wrist behave treat actual wasp year salad speed social layer crew genius"
+    name: test1
+    coins: ["20000snr", "200000000stake"]
+    # address: snr1cyyzpxplxdzkeea7kwsydadg87357qnaglp5wk
+validator:
+  name: validator
+  staked: "70000000stake"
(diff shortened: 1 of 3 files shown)


test-config.yaml
```diff
@@ -0,0 +1,21 @@
+accounts:
+  - mnemonic: "satisfy adjust timber high purchase tuition stool faith fine install that you unaware feed domain license impose boss human eager hat rent enjoy dawn"
+    name: validator
+    coins: ["20000snr", "100000000stake"]
+    # address: snr1phaxpevm5wecex2jyaqty2a4v02qj7qmqe4lvm
+  - mnemonic: "notice oak worry limit wrap speak medal online prefer cluster roof addict wrist behave treat actual wasp year salad speed social layer crew genius"
+    name: test1
+    coins: ["20000snr", "200000000stake"]
+    # address: snr1cyyzpxplxdzkeea7kwsydadg87357qnaglp5wk
```
ntindle, line 9: I think Brayden and Josh talked during our demo about swapping out account types. Did that happen?

prdlk: is putting mnemonic in config safe? could this be an env variable in github? @lakshya-sky @ntindle

lakshya-sky: @ntindle This accounts are needed to run the cosmos chain with the same address during restarts, I don't think we can do MPC related stuff in here.
@prnk28 This are completely safe, since we don't use them in production, also these are taken from Terra dev configs.

ntindle: @mcjcloud

mcjcloud: I think this is fine for now, once we get MPC figured out better we can come back and refactor this.

mcjcloud (approved):

joshLong145 (commented):
Approving but we should use a non root user

joshLong145 (approved):

PR #636, opened 2022-09-28
Extract hardcoded config values from code to test environment config file

### LINKED ISSUE #621: Extract hardcoded config values from code to test environment config file

Oozyx:
## Changes

- Introduce **godotenv** package, which allows for setting environment variables in a **.env** file
- Extract hardcoded environment variables from **pkg/client/client.go** and place in **.env** file
- Provide an **.env.example** file for reference
- Update readme with test instructions to setup test environment
- Add **internal/projectpath/path.go** to find project root path. To be used with **.env**

## Checklist

- [ ] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 


joshLong145 (changes requested):
It seems this wont work if the`.env` file is not available. please add `ldflags` for each for the `beta` address to the `scripts/bind.sh` for all targets.

ntindle:
> It seems this wont work if the`.env` file is not available. please add `ldflags` for each for the `beta` address to the `scripts/bind.sh` for all targets.

Can these flag be overridden at runtime?

joshLong145 (changes requested):
looks like the generated `request_pb.go` was not committed to this branch. could you add it so ci can pick up the new enum and properties?

pkg/client/client.go
```diff
@@ -1,33 +1,91 @@
...
+	BLOCKCHAIN_RPC_BETA    = "137.184.190.146:9090"
+
+	// -- Services --
+	IPFS_ADDRESS      = "https://ipfs.sonr.ws"
+	IPFS_API_ADDRESS  = "https://api.ipfs.sonr.ws"
+	VAULT_API_ADDRESS = "http://164.92.99.233"
+)
+
+type Client struct {
+	clientMode mt.ClientMode
```
ntindle, line 37: GHA thinks this is undefined. `v1.ClientMode` Can you check?

joshLong145 (approved):

ntindle (approved):

pkg/vault/create.go
```diff
@@ -80,8 +85,24 @@ func (v *vaultImpl) CreateVault(d string, deviceShards [][]byte, dscPub string,
...
+	env_path := filepath.Join(projectpath.Root, ".env")
+
+	var uri string
+	// by default use .env if it exists
+	_, err = os.Stat(env_path)
+	if errors.Is(err, os.ErrNotExist) {
+		uri = "https://vault.sonr.ws"
+	} else {
+		err = godotenv.Load(env_path)
+		if err != nil {
```
mcjcloud: If there's an error at this stage, `uri` is unlikely to get a value, so it should default still to `https://vault.sonr.ws`

pkg/vault/vault.go
```diff
@@ -16,8 +23,23 @@ type vaultImpl struct {
 }
 
 func New() VaultClient {
+	env_path := filepath.Join(projectpath.Root, ".env")
```
mcjcloud, line 26: This code is duplicated between here and `create.go`. Consider moving it to its own function.

mcjcloud (changes requested):

pkg/vault/create.go
Oozyx: I added a function that gets the uri, and instead of exiting with error I simply log the error and return a default value for uri.

mcjcloud (approved):

pkg/motor/motor.go
```diff
@@ -83,7 +85,7 @@ func EmptyMotor(r *mt.InitializeRequest, cb common.MotorCallback) (*motorNodeImp
 func initMotor(mtr *motorNodeImpl, options ...mpc.WalletOption) (err error) {
 	mtr.log = logger.New(mtr.logLevel, "motor")
 	// Create Client instance
-	mtr.Cosmos = client.NewClient(client.ConnEndpointType_BETA)
+	mtr.Cosmos = client.NewClient(mtr.clientMode)
```
ntindle, line 88: @mcjcloud

PR #654, opened 2022-09-30
Add CreateBucket method

prdlk:
## Changes

* Item 1
* Item 2

## Checklist

* [ ] Unit tests
* [ ] Documentation

## References *(optional)*

Fixes
Connects


prdlk:
Added create bucket method to bind

pkg/motor/interface.go
```diff
@@ -50,7 +50,7 @@ type MotorNode interface {
 	// Buckets
 
 	// Creates a new bucket with the defined properties in the request. Returns and instance of `bucket`. before returning both content and buckets are resolved.
-	CreateBucket(context.Context, mt.CreateBucketRequest) (bucket.Bucket, error)
+	CreateBucket(mt.CreateBucketRequest) (*mt.CreateBucketResponse, bucket.Bucket, error)
```
ntindle, line 54: Will this one break speedway?

pkg/motor/create_bucket.go
```diff
@@ -13,41 +12,41 @@ import (
 	bt "github.com/sonr-io/sonr/x/bucket/types"
 )
 
-func (mtr *motorNodeImpl) CreateBucket(ctx context.Context, request mt.CreateBucketRequest) (bucket.Bucket, error) {
```
ntindle, line 16: Will this change break speedway

mcjcloud: It'll just require changes next time speedway upgrades to this version

pkg/motor/create_bucket.go
```diff
@@ -59,5 +58,7 @@ func (mtr *motorNodeImpl) CreateBucket(ctx context.Context, request mt.CreateBuc
 
 	mtr.AddBucketServiceEndpoint(mtr.GetClient().GetRPCAddress(), cbresp.WhereIs.Did)
 
-	return b, nil
+	return &mt.CreateBucketResponse{
+		Did: cbresp.WhereIs.Did,
```
mcjcloud: Would it be useful to include the `WhereIs` here as well?

mcjcloud (approved):

pkg/motor/create_bucket.go
joshLong145: Yea, speedway won't break if it doesn't update to it.

joshLong145 (approved):

pkg/motor/create_bucket.go
prdlk: Yeah imma update the response proto to include and merge @mcjcloud

pkg/motor/interface.go
prdlk: The motor impl itself has a context, IDK why we're passing it into methods

PR #664, opened 2022-10-06
Login with alias

### LINKED ISSUE #601: Add Login with Alias to Motor Node

mcjcloud:
Adds the ability to login using an alias

## Changes

- Initialize motor client before `initMotor` in order to query WhoIs
- Rename `Did` to `AccountId` in login requests

Closes #601 #602 


pkg/motor/login.go
```diff
@@ -31,7 +34,7 @@ func (mtr *motorNodeImpl) Login(request mt.LoginRequest) (mt.LoginResponse, erro
 	}
 
 	return mtr.LoginWithKeys(mt.LoginWithKeysRequest{
-		Did:       request.Did,
+		AccountId: request.AccountId,
```
joshLong145, line 37: can this `AccountId` be a `DID` or an `Alias`?

pkg/motor/login.go
```diff
@@ -40,15 +43,38 @@ func (mtr *motorNodeImpl) Login(request mt.LoginRequest) (mt.LoginResponse, erro
...
+		whoIs = whoIsResp.WhoIs
+	} else {
+		whoIsResp, err := mtr.QueryWhoIs(mt.QueryWhoIsRequest{
+			Did: request.AccountId,
+		})
+		if err != nil {
+			return mt.LoginResponse{}, fmt.Errorf("query WhoIs: %s", err)
+		}
+		whoIs = whoIsResp.WhoIs
+	}
```
joshLong145, line 71: i think we could simplify this if we kept `DID` on the account struct and added `AccountId` as a seperate property. eliminates the need to string parse.

joshLong145 (changes requested):

pkg/motor/create_account.go
```diff
@@ -46,6 +46,9 @@ func (mtr *motorNodeImpl) CreateAccount(request mt.CreateAccountRequest) (mt.Cre
 
 // CreateAccountWithKeys allows PSK and DSC to be provided manually
 func (mtr *motorNodeImpl) CreateAccountWithKeys(request mt.CreateAccountWithKeysRequest) (mt.CreateAccountWithKeysResponse, error) {
+	// Create Client instance
+	mtr.Cosmos = client.NewClient(client.ConnEndpointType_BETA)
```
ntindle: ?

pkg/motor/login.go
```diff
@@ -40,15 +43,38 @@ func (mtr *motorNodeImpl) Login(request mt.LoginRequest) (mt.LoginResponse, erro
...
 func (mtr *motorNodeImpl) LoginWithKeys(request mt.LoginWithKeysRequest) (mt.LoginResponse, error) {
-	if request.Did == "" {
+	if request.AccountId == "" {
 		return mt.LoginResponse{}, fmt.Errorf("did must be provided")
 	}
 
-	mtr.Address = request.Did
+	// Create Client instance
+	mtr.Cosmos = client.NewClient(client.ConnEndpointType_BETA)
+
```
ntindle, line 52: I think webisoft built out a wrapper for this to grab the same one across the entire tool chain. Also shouldn’t this be wrapped in a singleton?

pkg/motor/registry_test.go
```diff
@@ -150,19 +150,45 @@ func Test_LoginAndMakeRequest(t *testing.T) {
 		return
 	}
 
-	req := mt.LoginRequest{
-		Did:      ADDR,
-		Password: "password123",
+	req := mt.LoginWithKeysRequest{
+		AccountId: ADDR,
```
ntindle: Shouldn’t this use the DSC not password?

pkg/motor/login.go
mcjcloud: Yes.

pkg/motor/login.go
mcjcloud: String parsing is currently what's being done in Flutter so I'm copying that logic here.

I think having it as one field simplifies the interface for the developer, because in most cases this value is going to come from a single text field or something.

pkg/motor/create_account.go
mcjcloud: This has to be done before `initMotor` now so that the client is ready to potentially query the address.

pkg/motor/registry_test.go
mcjcloud: It doesn't really make a difference, the login itself isn't being tested here.

pkg/motor/login.go
mcjcloud: Where is the wrapper?

pkg/motor/login.go
ntindle: I saw where you’re coming from josh and I think I agree with Brayden. Otherwise what happens if you pass conflicting information?

pkg/motor/login.go
ntindle: https://github.com/sonr-io/sonr/pull/636/files#r989477398

ntindle: This is where I thought it was used. Let me know if it is wrong and also if it should be added to the backlog

mcjcloud: Fixed

ntindle (approved):

joshLong145 (approved):

PR #669, opened 2022-10-07
Add crisis module

lakshya-sky:
### Adds:
- Adds crisis invariants for bucket and registry modules.
- Generates Events for transactions which we can use later for off-chain computation.

### Invariant Commands:
- `sonrd tx crisis invariant-broken bucket total-where-is --from=alice --chain-id=sonrd`
- `sonrd tx crisis invariant-broken registry single-alias-owner --from=alice --chain-id=sonrd`
- `sonrd tx crisis invariant-broken registry nonnegative-sell-price --from=alice --chain-id=sonrd`


ntindle:
@lakshya-sky what is this waiting on?

docker/Dockerfile.localnet
```diff
@@ -8,9 +8,13 @@ RUN mv /app/test-config.yaml /app/config.yml
...
 USER tendermint
+RUN /bin/bash -c 'ARCH=`uname -m` && \
+    if [ "$ARCH" == "aarch64" ]; then export ARCH="arm64"; else export ARCH="amd64"; fi && \
+    export URL=https://github.com/TomWright/dasel/releases/latest/download/dasel_linux_$ARCH && \
+    wget $URL -O /go/bin/dasel && chmod +x /go/bin/dasel'
 
-RUN go install github.com/tomwright/dasel/cmd/dasel@master
-RUN ignite chain init
+#RUN go install github.com/tomwright/dasel/cmd/dasel@master
+RUN /bin/bash -c 'ignite chain init --skip-proto -y'
```
ntindle: Why? This is needed for local dev I'd assume

x/bucket/types/events.go
```diff
@@ -0,0 +1,14 @@
...
+	EventTypeCreateWhereIs = "createWhereIs"
+	EventTypeUpdateWhereIs = "updateWhereIs"
+	EventTypeDeleteWhereIs = "deleteWhereIs"
+
+	AttributeValueCategory = ModuleName
+	AttributeKeyTxType     = "txType"
+
+	AttributeKeyCreator = "whereIsCreator"
+	AttributeKeyDID     = "did"
+	AttributeKeyLabel   = "lable"
```
ntindle: label

ntindle (commented):
I don't really know enough about this to know if this is done correctly. Would love help from Webisoft to review here

docker/Dockerfile.localnet
lakshya-sky: I was facing longer wait times for building containers. So I changed a bit things
1. download dasel binary instead on building it
2. Skip proto generation
3. the latest ignite(v0.24.0) image requires to use v1 config which is different than we use right now. `-y` will convert current `test-config.yml` to v1 format. Since this is happening inside container this wouldn't affect code base on our disk.

x/bucket/types/events.go
lakshya-sky: lol

ntindle:
Is this ready?

mcjcloud (approved):

prdlk (approved):

PR #685, opened 2022-10-11
Fix pkg/client tests and make them CI Run-able

Oozyx:
## Changes

- Add client test suite 
- Add tests for client address getters


pkg/motor/motor.go
```diff
@@ -274,3 +276,83 @@ func (w *motorNodeImpl) SendTx(routeUrl string, msg sdk.Msg) ([]byte, error) {
 	}
 	return resp.GetTxResponse().Marshal()
 }
+
+func SetupTestAddressWithKeys(motor MotorNode) (error) {
```
mcjcloud: I don't think this is a good place for these test-related functions.

pkg/motor/motor.go
```diff
@@ -274,3 +276,83 @@ func (w *motorNodeImpl) SendTx(routeUrl string, msg sdk.Msg) ([]byte, error) {
...
+	name := fmt.Sprintf("./test_keys/%s", n)
+	file, err := os.Create(name)
+	if err != nil {
+		return false
+	}
+	defer file.Close()
+
+	_, err = file.Write(aesKey)
+	return err == nil
+}
```
mcjcloud: I notice there are a lot of missing newlines in files. Are you using gofmt?

mcjcloud (changes requested):

pkg/client/client_test.go
```diff
@@ -1,44 +1,78 @@
...
+func (suite *ClientTestSuite) Test_GetFaucetAddress() {
+	env_path := filepath.Join(projectpath.Root, ".env")
+	err := godotenv.Load(env_path)
+  assert.NoError(suite.T(), err, "Load .env success")
+	assert.Equal(suite.T(), os.Getenv("BLOCKCHAIN_FAUCET"), suite.motorNode.GetClient().GetFaucetAddress(), "Faucet address should be the same")
+}
+
+func (suite *ClientTestSuite) Test_GetRPCAddress() {
+	env_path := filepath.Join(projectpath.Root, ".env")
+	err := godotenv.Load(env_path)
```
joshLong145: do we have to do this for every test? i think it could be in the setup.

joshLong145 (changes requested):

pkg/motor/motor.go
joshLong145: agreed should be in `testutils`

pkg/client/client_test.go
Oozyx: Now it's in the suite setup

pkg/motor/motor.go
Oozyx: my vscode is set to formatOnSave but for some reason newlines aren't properly handled. I just ran go fmt on the package so should be fine now

pkg/motor/motor.go
Oozyx: refactored

pkg/client/client.go
```diff
@@ -145,9 +145,9 @@ func (c *Client) GetIPFSApiAddress() string {
 	}
 
 	err = godotenv.Load(env_path)
-  if err != nil {
-    log.Fatal(err)
-  }
+	if err != nil {
+		log.Fatal(err)
```
joshLong145, line 149: why two tabs on these?

joshLong145 (approved):
Fix formatting errors with double tabs and its good to merge

ntindle:
Can we fix `app/simulation_test.go` and `testutil/network/network.go` as well in this pr? They are throwing errors on every pr right now and it should be a pretty quick fix

mcjcloud:
Where are the simulation test changes coming from? Are they relevant to this PR?

mcjcloud (approved):

PR #692, opened 2022-10-12
fixed issues related to document upload

mcjcloud:
Fixes issue in upload document

### DIFF
--- a/pkg/motor/document_builder.go
+++ b/pkg/motor/document_builder.go
@@ -26,7 +26,7 @@ func (mtr *motorNodeImpl) NewDocumentBuilder(did string) (*document.DocumentBuil
 func (mtr *motorNodeImpl) GetDocument(req mt.GetDocumentRequest) (*mt.GetDocumentResponse, error) {
 	dag, err := mtr.queryDocument(req.GetCid())
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/motor/document_builder.go
```diff
@@ -64,9 +76,19 @@ func (mtr *motorNodeImpl) UploadDocument(req mt.UploadDocumentRequest) (*mt.Uplo
 		return nil, err
 	}
 
+	// Normalize number values
+	// json.Unmarshal decodes all numbers as float64 by default
+	for _, f := range builder.GetSchema().GetFields() {
+		if f.GetKind() == st.Kind_INT {
+			doc[f.Name] = int(doc[f.Name].(float64))
```
ntindle, line 83: Does float64 work in flutter @prnk28

mcjcloud: This is a json encoding thing, so no matter what format it is in flutter, it will come out as `float64` in Go land.

mcjcloud: And the DAG gets converted into a SchemaDocument in Go, which is proto serialized when sent back to flutter.

ntindle (approved):

joshLong145 (approved):

PR #695, opened 2022-10-13
Add CID field back to SchemaDocument

prdlk:
This fixes the Bucket interface in Motor Flutter

## Changes

* Remove auto push to buf.build in makefile
* Add CID field to SchemaDocument
* Comment out broken tests

Fixes
Connects


joshLong145:
why do we need the`cid` referenced in the document? It will only be available after `uploading`

PR #702, opened 2022-10-17
Upgrade module

UvRoxx:
## Changes

- Added Upgrade Module and Handler for managing the chains state upgrade.
- Added Cosmovisor setup for auto-downloading and updating binary with respect to on chain governance.

## API Updates

-  Query method for upgrade plan and info added

### New Features *(required)*

- This pull request adds chain upgrade capabilities as well as scripts for deploying and managing cosmovisor.

### Deprecations *(required)*
- None
### Enhancements *(optional)*
- Addition of upgrade.go and integration of upgrade handler in app.go
-
## Checklist

- [ ] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 


app/upgrades.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	store "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+)
+
+// next upgrade name
+const upgradeName = "v2"
```
joshLong145, line 13: nit: should be config driven

scripts/propose_upgrade.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "Enter From User Address:"
+read -r from
+echo "Submit Proposal To Upgrade";
+# Submit Proposal
+sonrd tx gov submit-proposal software-upgrade --upgrade-height $height --from $from --yes --title $title --description test upgrade
+echo "Enter the Proposal ID:"
+read -r proposal_id
+# Deposit For Proposal
+echo "Depositing Funds For Proposal ID: $proposal_id"
+sonrd tx gov deposit $proposal_id 10000000stake --from $from --yes
```
joshLong145, line 16: should we make stake  configurable?

scripts/propose_upgrade.sh
```diff
@@ -0,0 +1,19 @@
...
+# Vote for the Proposal
+echo "Voting Yes For Proposal ID: $proposal_id"
+sonrd tx gov vote $proposal_id yes --from $from --yes
```
joshLong145, line 19: do we always want to vote yes on the upgrade?

joshLong145 (changes requested):

scripts/propose_upgrade.sh
utkush: This is configured in the apps genesis.Json along with voting time.

scripts/propose_upgrade.sh
utkush: Yes, we need to Vote yes every time, in a production scenario the upgrade requires a majority of the network to vote Yes in order to pass the proposal. Since the upgrade is a gov proposal based.

app/upgrades.go
utkush: This name is for the binary name the upgrade represents, it needs to be compiled in the chain binary, so I don't think we can move it to config.

joshLong145: we could move it out to a config that sets the version. if this is left as it is we will have to manually bump this for every new release.

joshLong145: it could be configured with viper in `sonrd` and loaded through env.

scripts/propose_upgrade.sh
joshLong145: what if a validator wants to vote no on an upgrade?

UvRoxx: Well then the upgrade technically wouldn't pass, it needs a majority of the network to agree for it to be passed just like any other governance proposal.

mcjcloud: In what contexts does this script run? Are we forcing people to vote yes on proposal inadvertently?

UvRoxx: this script was meant to be used for the demo and if a validator wishes to approve a proposal they need to execute the last line from it and vote yes to the proposal for the upgrade for it to pass. If a majority of the validators agree to a proposal that proposal is automatically passed.

mcjcloud: Okay sgtm.

app/app.go
```diff
@@ -769,6 +773,15 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
 }
 
+// RegisterUpgradeHandlers returns upgrade handlers
```
mcjcloud: Is this code needed?

app/upgrades.go
```diff
@@ -0,0 +1,33 @@
...
+
+// RegisterUpgradeHandlers returns upgrade handlers
+func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
+	app.UpgradeKeeper.SetUpgradeHandler(upgradeName, func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		return app.mm.RunMigrations(ctx, cfg, vm)
+	})
+
+	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
+	if err != nil {
+		panic(fmt.Sprintf("failed to read upgrade info from disk %s", err))
```
mcjcloud, line 23: Avoid panicking if possible. Is there a more graceful way to fail?

UvRoxx: It panics cause this operations objective if it fails is to halt the chain, to prevent state inconsistencies across nodes.

prdlk (approved):

app/app.go
UvRoxx: Registering upgrade handler is essential here.

ntindle: I think hes referring to it being commented out?

ntindle:
Also looks like `bind` and `placid` lost their execute flags which could break ci

app/app.go
UvRoxx: Oh Yes that part I'll remove it since we move the handler to upgrade.go

ntindle (approved):

mcjcloud (approved):

joshLong145 (approved):

Issue #565, opened 2022-09-09
split bucket and schema keepers in chain to create stores based off `creator address`


joshLong145:
currently store keys are based off the object's `did` which creates a 1 - 1 ratio between the key and the value. which creates a single store for all data within the given module. this creates an issue with pagination where a single store is accessed and paged through. since the stores size is not an indication of the data size being returned. it makes paginating inaccurate.
To fix this, `bucket` and `schema` stores should key off `creator` and keep a store for each creator address. in order to keep pagination data accurate.

prdlk:
We should store objects like:
`did:snr:abc123/bucket-name/user-data-object-cid`

We need to create App specific WhoIs for App DID's for Schemas and buckets:
For Dev resources:
- Generate UDID for new application in registry `did:snr:<udid>
- Resolve schema from human readable name `did:snr:xyz789/schema/profile`

prdlk:
IPNS Time?? @ntindle @joshLong145 @mcjcloud @uncl3-b3nny @ma-sonr

PR #708, opened 2022-10-18
fix to bucket by creator query issues

joshLong145:
- fixes to bucket `queryByCreator` returning a list of buckets that where of the correct length, but all duplicate items.

## Changes
- updates to `QueryWhereIsByCreatorResponse` to no longer have a `nullable` option as it forces non pointer values for arrays where `[]types.WhereIs` and not `[]*types.WhereIs` 


pkg/motor/query.go
```diff
@@ -120,11 +120,11 @@ func (mtr *motorNodeImpl) QueryWhereIs(req mt.QueryWhereIsRequest) (*mt.QueryWhe
 }
 
 func (mtr *motorNodeImpl) QueryWhereIsByCreator(req mt.QueryWhereIsByCreatorRequest) (*mt.QueryWhereIsByCreatorResponse, error) {
-	resp, err := mtr.GetClient().QueryWhereIsByCreator(req.Creator, req.Pagination)
+	resp, err := mtr.GetClient().QueryWhereIsByCreator(req.Creator, nil)
```
mcjcloud: Is pagination broken? Is there an issue to track that?

mcjcloud (approved):

pkg/motor/query.go
joshLong145: its not accurate at this time due to key indexes in the keeper there is an issue for this: https://github.com/sonr-io/sonr/issues/565 i should put back the pagination instance if provided. as it will only effect requests that explicitly define a pagination limit.

prdlk:
These are out of scope additions that conflict with our parallel work - not approving

joshLong145:
> These are out of scope additions that conflict with our parallel work - not approving

this is stale and his been up for a long time closing as #723 connects this

PR #710, opened 2022-10-18
creation of test suite for `pkg/ipns`

### LINKED ISSUE #648: Fix pkg/ipns tests and make them CI Run-able
Needs an IPFS local shell

joshLong145:
- removes test case from `ipns` that would always fail due to the given `cid` not being present. 


ntindle:
Should this just upload something in a setup?

joshLong145:
> Should this just upload something in a setup?

the module isnt being used in any part of sonr at this time. when the module is brought in for creating persistent address it can be tested. this is being removed now to no longer fail the test suite.

joshLong145:
> Should this just upload something in a setup?

Added it to a suite so an `document` is created and uploaded with a cid

ntindle (approved):

ntindle:
Closes #648

mcjcloud (approved):

PR #712, opened 2022-10-19
Josh/eng 204 WalletCreateEvent

joshLong145:
- Addition of `WalletEvent` protobuf and `OnWalletEvent` on `MotorCallback`

## Changes
- Integrates `OnWalletEvent` into `CreateAccount` and `CreateAccountWithKeys`
- creation of `utils` for marshalling the `WalletEvent` and triggering the callback.

Connects: https://github.com/sonr-io/motor-flutter/pull/41


prdlk (approved):
LGTM! Pending bind

bind/motor-mobile/bind.go
```diff
@@ -16,15 +16,15 @@ var (
 	instance    mtr.MotorNode
 )
 
-func Init(buf []byte, cb ct.MotorCallback) ([]byte, error) {
+func Init(buf []byte) ([]byte, error) {
```
prdlk: This change would break Motor Bind @joshLong145 please dont include this change

pkg/motor/create_account.go
```diff
@@ -11,20 +11,42 @@ import (
...
 	"github.com/sonr-io/sonr/pkg/tx"
 	"github.com/sonr-io/sonr/pkg/vault"
+	"github.com/sonr-io/sonr/third_party/types/common"
 	mt "github.com/sonr-io/sonr/third_party/types/motor/api/v1"
 	rt "github.com/sonr-io/sonr/x/registry/types"
 )
 
 func (mtr *motorNodeImpl) CreateAccount(request mt.CreateAccountRequest) (mt.CreateAccountResponse, error) {
 	// create DSC and store it in keychain
+	mtr.triggerWalletCreateEvent(common.WalletEvent{
```
mcjcloud: We should call this method something other than `triggerWalletCreateEvent` if it's going to be used for any type of event. Like `triggerWalletEvent`.

pkg/motor/create_account.go
```diff
@@ -11,20 +11,42 @@ import (
...
+		Completed:    false,
+	})
 	dsc, err := kr.CreateDSC()
 	if err != nil {
+		mtr.triggerWalletCreateEvent(common.WalletEvent{
+			Type:         common.WALLET_EVENT_TYPE_KEY_CREATE_END,
+			HasErrored:   true,
+			ErrorMessage: err.Error(),
+			Message:      "",
+			Completed:    true,
```
mcjcloud: Does this mean the event is completed or the process it's around? It looks like you have an event type specifically denoting the "end" of this process, so the boolean here is redundant.

mcjcloud: I also think an error can be determined by the presence of `ErrorMessage`, and the `HasErrored` wouldn't be necessary

third_party/types/common/callback.go
```diff
@@ -13,10 +15,16 @@ func DefaultCallback() MotorCallback {
...
-// func (cb *defaultCallback) OnMotorEvent(msg string, isDone bool) {
-// 	log.Printf("message: %s done: %t", msg, isDone)
-// }
-
 func (cb *defaultCallback) OnDiscover(data []byte) {
-	log.Println("ERROR: MotorCallback not implemented.")
+	log.Println(ErrDefaultStillImplemented.Error())
+}
+
+func (cb *defaultCallback) OnWalletEvent(data []byte) {
```
mcjcloud, line 22: Can these changes be unit tested?

third_party/types/common/errors.go
```diff
@@ -29,4 +29,5 @@ var (
 	ErrListenerRequired          = errors.New("listener was not Provided")
 	ErrMDNSInvalidConn           = errors.New("invalid Connection, cannot begin MDNS Service")
 	ErrMotorWalletNotInitialized = errors.New("motor Wallet is not initialized")
+	ErrDefaultStillImplemented   = errors.New("default implementation of event callback still defined")
```
mcjcloud: I think the change in this message makes it less clear that there is a problem. If the default implementation is being used, I think it's fair to error with "callback not implemented"

mcjcloud (changes requested):

prdlk:
Call back shouldn't be removed from Init method, this breaks the platform channel and callback isn't passed @joshLong145

bind/motor-mobile/bind.go
joshLong145: I just noticed this after build the framework. i reverted that change

pkg/motor/create_account.go
joshLong145: I can rename to this. its only called in `create_account` methods so i named it like that.

pkg/motor/create_account.go
joshLong145: that makes sense, the completion flags are a bit redundant. ill remove them

third_party/types/common/errors.go
joshLong145: thats fair, ill change this.

third_party/types/common/callback.go
joshLong145: we can unit test the implementation of the interface. and use the utility methods

mcjcloud (approved):
lgtm

PR #715, opened 2022-10-21
addition of callback tests

joshLong145:
- addition of callback tests for motor
## Changes

- Item 1
- Item 2

## Checklist

- [ ] Unit tests
- [ ] Documentation

## References *(optional)*

Fixes 
Connects 


pkg/motor/callback_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (tc *testCallback) ClearState() {
+	tc.WalletEventExec = false
+}
+
+func (tc *testCallback) GetWalletEventState() bool {
+	return tc.WalletEventExec
+}
+
+func Test_Callbacks(t *testing.T) {
```
mcjcloud, line 29: Can this use the test suite?

joshLong145: it could but then it would be slower to run since it would have to go through all the account creation steps. these tests dont depend on such operations so it was isolated.

mcjcloud: Will this be included in CI?

ntindle: Yeah +1 really want every test to be run-able in CI

joshLong145: this is ci runnable and will run alongside the motor test suite. we can add it to the suite but it doesnt need the overhead contained in it.

mcjcloud: Sgtm

6missedcalls (approved):
LGTM

mcjcloud (approved):

PR #717, opened 2022-10-24
Merge Pradsnr/blok 13

prdlk:
Rescaffolded the Bucket module to allow for upgraded changes. Note - this is simply for scaffolding the skeleton module, implementation of methods and connecting ipfs are separate tasks.

## Changes

* See Design Doc on Linear for Reference


lakshya-sky:
Looks okay. But how do we maintain access control and visibility? since everything we store on the block-chain could be retrieved one way or another.

lakshya-sky (approved):

ntindle (approved):
Approved without review. Theres a huge amount of changes here and it fails a large number of CI and tests

prdlk:
> Looks okay. But how do we maintain access control and visibility? since everything we store on the block-chain could be retrieved one way or another.

Checkout https://docs.textile.io/buckets/

prdlk:
> Approved without review. Theres a huge amount of changes here and it fails a large number of CI and tests

See the error messages from CI Tests, its because the type names have changed and accompanying tests arent written.

PR #718, opened 2022-10-25
Pradsnr/blok 74

prdlk:
Created internal logic to fetch content from new bucket mechanism in `x/bucket/internal`.

## Changes

* `DownloadBucket` - Recursively downloads all bucket items
* `DownloadBucketAsync` - Downloads all bucket items in goroutines and returns a ProgressHandlerCallback
* `LsBucket` - Lists all the items within the bucket
* `ReadBucketItem` - Reads an item given its name from the bucket
* `WriteBuckettem` - Writes a new bucket item to IPFS

### New Features *(required)*

* Bucket Item Interface
* SchemaDocument has a conversion method to become a BucketItem

### Deprecations *(required)*

* `internal/bucket`

Fixes
Connects


prdlk:
This merge is depending on the rescaffolding one

lakshya-sky (approved):

ntindle (approved):

ntindle:
Approved without review.

prdlk:
> Approved without review.

This one probably needs a review

ntindle:
Merge these two PRs into a feature branch until all the motor etc changes are ready to go IMO

proto/bucket/bucket.proto
```diff
@@ -19,5 +17,38 @@ message BucketContent {
...
+  // Name of the new bucket.
+  string name = 3;
+
+  // Visibility of the new bucket.
+  BucketVisibility visibility = 4;
+
+  // IsActive flag of the new bucket.
+  bool is_active = 6;
+
+  // Timestamp of the new bucket.
```
joshLong145: is size the total size of the directory? comment is left over

x/bucket/internal/ipfs_download_bucket.go
```diff
@@ -0,0 +1,125 @@
...
+
+// This method is called on a routine and updates the array of items and current progress until the download is complete
+func (cb *bucketDownloadProgressCallback) handleChannels() {
+	for {
+		select {
+		case item := <-cb.itemsChan:
+			cb.items = append(cb.items, item)
+		case progress := <-cb.progressChan:
+			cb.current = progress
+			if cb.current == cb.total {
```
joshLong145, line 101: we should check if current has gone over the total as well or this will never complete

x/bucket/internal/ipfs_generate_bucket.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func GenerateBucket(sh *shell.Shell, b *bt.Bucket, address string) (*did.Service, error) {
+	path := b.GetPath(address)
+	service := b.GetDidService(address)
+	err := sh.FilesMkdir(context.Background(), path, shell.FilesWrite.Create(true))
+	if err != nil {
+		return service, fmt.Errorf("Failed to Make Directory %s in IPFS Node, %e", path, err)
+	}
+	res, err := sh.FilesStat(context.Background(), path)
```
joshLong145, line 19: we should pass `context` from the caller so if the operation needs to be canceled it can from outer scope. This applies to all methods in `x/bucket/internal`

x/bucket/internal/ipfs_write_bucket_item.go
```diff
@@ -0,0 +1,19 @@
...
+	"context"
+	"fmt"
+
+	shell "github.com/ipfs/go-ipfs-api"
+	bt "github.com/sonr-io/sonr/x/bucket/types"
+)
+
+func WriteBucketItem(sh *shell.Shell, b *bt.Bucket, address string, item bt.BucketItem) (string, error) {
+	itemPath := b.GetPath(address, item.Name())
+	err := sh.FilesWrite(context.Background(), itemPath, bytes.NewReader(item.Content()), shell.FilesWrite.Create(true))
```
joshLong145, line 14: standardiznig the encoding on the `item.Content` will help with retrieval

x/bucket/types/bucket.go
```diff
@@ -0,0 +1,78 @@
...
+	fmt "fmt"
+	io "io"
+	"strings"
+
+	"github.com/sonr-io/sonr/pkg/did"
+	"github.com/sonr-io/sonr/pkg/did/ssi"
+)
+
+func (b *Bucket) GetPath(address string, segments ...string) string {
+	path := fmt.Sprintf("/%s/%s", address, b.Uuid)
```
joshLong145, line 13: it seems we are creating a top level directory structure of `address/bucket-id` does this get created when the bucket is created?

joshLong145 (approved):
if we are to use `IPNS` I'm unsure of how we will leverage an asymmetric key pair to produce the addresses. we might be able to use the remote node's key pairs and allow it to sign but I'm unsure of the path forward.

x/bucket/internal/ipfs_generate_bucket.go
```diff
@@ -0,0 +1,28 @@
...
+
+	shell "github.com/ipfs/go-ipfs-api"
+	"github.com/sonr-io/sonr/pkg/did"
+	bt "github.com/sonr-io/sonr/x/bucket/types"
+)
+
+func GenerateBucket(sh *shell.Shell, b *bt.Bucket, address string) (*did.Service, error) {
+	path := b.GetPath(address)
+	service := b.GetDidService(address)
+	err := sh.FilesMkdir(context.Background(), path, shell.FilesWrite.Create(true))
```
joshLong145, line 15: i think this will remake the top level `address` directory as well if a bucket has already been created with this address.

x/bucket/types/bucket.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, item := range items {
+		path = fmt.Sprintf("%s/%s", path, item)
+	}
+	return path
+}
+
+type bucketItemImpl struct {
+	name    string
+	content []byte
+}
```
joshLong145, line 43: does this mean buckets cannot reference other buckets? content is now an arbitrary `byte` array

x/bucket/internal/ipfs_write_bucket_item.go
prdlk: SchemaDocuments only for now? @joshLong145

x/bucket/types/bucket.go
prdlk: Yes see generate bucket @joshLong145

PR #721, opened 2022-10-26
Block 69

UvRoxx:
## Changes

- Added Custom Decorator for Signature Verification
- Added Method to Perform Song MPC Signature Verification along with Standard verification.

## API Updates

- None

### New Features *(required)*

- Signature Verification is now enabled for Sonr MPC as well as Standard Cosmos Wallet Signatures
### Enhancements *(optional)*

- None

## Checklist

- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/app/sign_verify.go
+++ b/app/sign_verify.go
@@ -83,7 +83,7 @@ func (svd SigVerificationDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
 				if snrErr := VerifySonrSignature(pubKey, signerData, sig.Data, svd.signModeHandler, tx); snrErr == nil {
 					return next(ctx, tx, simulate)
 				} else {
-					fmt.Printf("ethereum signature verification failed; %s", snrErr.Error())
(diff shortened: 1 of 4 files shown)


app/sign_verify.go
```diff
@@ -0,0 +1,143 @@
...
+			accNum = acc.GetAccountNumber()
+		}
+		signerData := authsigning.SignerData{
+			ChainID:       chainID,
+			AccountNumber: accNum,
+			Sequence:      acc.GetSequence(),
+		}
+
+		// no need to verify signatures on recheck tx
+		if !simulate && !ctx.IsReCheckTx() {
```
joshLong145, line 83: would we want this to be an or?

joshLong145 (approved):

prdlk (approved):

app/sign_verify.go
mcjcloud: This is right. "If not simulate and not recheck txn, then verify signature". So if either is true, verification won't happen

PR #722, opened 2022-10-26
Josh/blok 27

joshLong145:
`did verification method` - Updated to use a `jwk` based on the device address to generate `edcsa` keys
`new key proto in wallet` - New key generation from wallet public address to create asymmetric `edcsa` keys
`new jwk impl in verification` - Now creates a sinage key which holds the public, private `edcsa` pair
`start to did resolve impl` - basic implementation of did resolve, currently missing some metadata info needed to comply with `did properties`.

## Changes
`Change to verification method` vm is now a `JSONWebKey2020` with the derived `edcsa` key

## Checklist

- [x] Unit tests
- [ ] Documentation


pkg/crypto/jwx/jws.go
```diff
@@ -4,9 +4,9 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jws"
 )
 
-func (x *jwxImpl) Sign(payload []byte, opts ...jws.SignOption) ([]byte, error) {
+func (x *jwxImpl) Sign(payload []byte, key interface{}, opts ...jws.SignOption) ([]byte, error) {
```
mcjcloud, line 7: Why add the key as an argument instead of using the one on the struct?

pkg/crypto/mpc/wallet.go
```diff
@@ -1,6 +1,9 @@
 package mpc
 
 import (
+	"bytes"
+	std_edcsa "crypto/ecdsa"
```
mcjcloud: nit: should be "ecdsa" instead of "edcsa"

pkg/crypto/mpc/wallet.go
```diff
@@ -128,6 +131,27 @@ func (w *Wallet) PublicKeyProto() (*secp256k1.PubKey, error) {
...
+		return nil, err
+	}
+
+	key, err := std_edcsa.GenerateKey(elliptic.P256(), bytes.NewReader(p))
+	if err != nil {
+		return nil, err
+	}
+
+	return key, nil
+}
```
mcjcloud: Creating a new keypair from an existing public key isn't secure. Anybody could recreate this key with knowledge of the public key.

pkg/motor/create_account.go
```diff
@@ -105,7 +105,15 @@ func (mtr *motorNodeImpl) CreateAccountWithKeys(request mt.CreateAccountWithKeys
...
-	vm, err := did.NewVerificationMethodFromBytes(doc.GetID(), ssi.ECDSASECP256K1VerificationKey2019, *controller, mtr.GetPubKey().Bytes())
+	pk, err := mtr.Wallet.CreateEcdsaFromPublicKey()
+	if err != nil {
+		mtr.triggerWalletEvent(common.WalletEvent{
+			Type:         common.WALLET_EVENT_TYPE_DID_DOCUMENT_CREATE_ERROR,
+			ErrorMessage: err.Error(),
+		})
+		return mt.CreateAccountWithKeysResponse{}, err
+	}
+	vm, err := did.NewVerificationMethod(doc.GetID(), ssi.JsonWebKey2020, *controller, pk)
```
mcjcloud: Looks like you're replacing the public key in the verification method with a private key which doesn't match the spec and may present security issues.

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/sonr-io/sonr/pkg/client"
+	"github.com/sonr-io/sonr/pkg/crypto/mpc"
+	"github.com/sonr-io/sonr/pkg/did"
+)
+
+type ResolverImpl struct {
+	wallet *mpc.Wallet
+	client *client.Client
+	did.Resolver
```
mcjcloud: This is an interface, making a struct inherit from it will suppress any errors for unimplemented methods.

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
+package resolver
```
mcjcloud: Putting this package inside the `did` package would make more sense imo. "Resolver" lacks context

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+	did.Resolver
+}
+
+func New() did.Resolver {
+	return &ResolverImpl{}
+}
+
+// this could be abstracted into a provider for connections.
+// right now binded to the client pkg
+func (res *ResolverImpl) WithClient(client *client.Client) {
```
mcjcloud: nit: `WithClient` makes me think this method will return a `ResolverImpl`. `SetClient` would be more clear

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+	res.client = client
+}
+
+func (res *ResolverImpl) WithWallet(wallet *mpc.Wallet) {
+	res.wallet = wallet
+}
+
+func (r *ResolverImpl) Resolve(inputDID string) (*did.Document, *did.DocumentMetadata, error) {
+	if r.client == nil {
```
mcjcloud: If this struct needs a client to be useful, consider adding it to the `New` function to make it required.

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+	doc, err := res.DidDocument.ToPkgDoc()
+
+	if err != nil {
+		return nil, nil, err
+	}
+	t := time.Unix(res.Timestamp, 0)
+	md := make(map[string]interface{})
+
+	md["updated"] = res.Timestamp
+	md["deactivated"] = !res.IsActive
```
mcjcloud: Resolving a DID Document shouldn't update these properties, no changes are being made.

pkg/resolver/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+
+	metaData := did.DocumentMetadata{
```
mcjcloud: Should the metadata not exist on the DID Document itself?

mcjcloud (changes requested):

pkg/crypto/mpc/wallet.go
joshLong145: the intent is to create a key compatible with json web keys for signing this replaces the base58 public key such that signing of messages with the verification method is easier.

pkg/motor/create_account.go
joshLong145: this will pull. the public key out of the `privateKey` and use it to verify when singing the private key is still required

mcjcloud:
> * Added a new method to the wallet interface that allows for creating an ecdsa key from
>   the public key of a mpc generated wallet. This is used in jwx when signing messages with
>   our private keys, which are stored as wallets. The JWK library requires us to pass it an ECDSA
>   object and not just bytes so we need this function to create one on demand using our pubkey bytes.

This is cool but a little inaccurate. The JWK library does not require an ECDSA object to be passed in.

pkg/crypto/jwx/jws.go
joshLong145: the struct needs to use the public key to generate and the private to sign.

pkg/crypto/mpc/wallet_test.go
```diff
@@ -92,3 +95,21 @@ func Test_MPCSignMessage(t *testing.T) {
 	deserializedSigVerified := sig.Verify(w.Config().PublicPoint(), m)
 	assert.True(t, deserializedSigVerified, "deserialized signature is verified")
 }
+
+func Test_MPCAddressDeriveSameEDCSA(t *testing.T) {
```
mcjcloud: nit: *ECDSA

pkg/crypto/mpc/aes.go
```diff
@@ -104,3 +107,16 @@ func NewAesKey() ([]byte, error) {
 
 	return key, nil
 }
+
+func NewEcdsaFromAes(aes []byte) (*ecdsa.PrivateKey, error) {
+	for len(aes) < 40 {
```
mcjcloud, line 113: Is there documentation that suggests this padding? Not sure if this is necessary.

pkg/crypto/mpc/wallet.go
```diff
@@ -128,6 +131,26 @@ func (w *Wallet) PublicKeyProto() (*secp256k1.PubKey, error) {
...
 
+func (w *Wallet) CreateEcdsaFromPublicKey() (*std_ecdsa.PrivateKey, error) {
+	p, err := w.PublicKey()
+
+	// need to pad the key to 40 bytes for
+	// ecdsa key generation which sizes its buffers from
+	for len(p) < 40 {
+		p = append(p, 0)
+	}
+	if err != nil {
```
mcjcloud: This check should come before the padding.

pkg/did/resolver/resolver.go
```diff
@@ -0,0 +1,124 @@
...
+	t := time.Unix(res.Timestamp, 0)
+
+	// TODO: this should not be manually mapped and instead be stored within the did registry.
+	// Will need to update registry module to also hold Document metadata
+	// need updated, deactivated, versionId, nextUpdate, nextVersionId
+	// see: https://www.w3.org/TR/did-spec-registries/#did-document-metadata
+	md := make(map[string]interface{})
+
+	md["updated"] = res.Timestamp
+	md["deactivated"] = !res.IsActive
```
mcjcloud, line 62: I don't think these properties should be set during a query.

pkg/motor/create_account.go
```diff
@@ -226,6 +234,15 @@ func createVault(mtr *motorNodeImpl, request mt.CreateAccountWithKeysRequest) {
 	// update DID Document
 	mtr.DIDDocument.AddService(vaultService)
 
+	if err != nil {
```
mcjcloud: `err` isn't being reassigned between now and the previous check, so this will always be false.

pkg/motor/keys.go
```diff
@@ -0,0 +1,65 @@
...
+
+	- []byte -> encrypted content by the jwk which is encrypted and returned
+
+	- error => returned if there is an error in generating keys and encrypting content, other values will be nil
+*/
+func (mtr *motorNodeImpl) EncryptContent(data []byte) ([]byte, []byte, error) {
+	x := jwx.New()
+	x.SetKey(mtr.encryptionKey)
+
+	_, err := x.CreateEncJWK()
```
mcjcloud, line 31: Is this deterministic?

pkg/motor/keys.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, nil, err
+	}
+
+	keyBytes, err := x.MarshallJSON()
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	encyptedKey, err := j.EncryptJWE(keyBytes)
```
mcjcloud, line 66: This key can be decrypted by anybody because the user's public key is used to generate `j`.

mcjcloud (changes requested):

pkg/crypto/mpc/aes.go
joshLong145: when inspecting the source of the keygen the buffer needs to be 40 bytes. this was found by stepping through when the key is generated. ill add some docs.

pkg/did/resolver/resolver.go
joshLong145: we need to return the document metadata as per the spec for `resolve` the link within the comments has a bit more explanation. there needs to be follow up tasks to add things to the registry in the WhatIs

pkg/motor/keys.go
joshLong145: this is true, i will generate a new key and it will have to be vault persisted or stored in the bucket encrypted by the encryption key

ntindle (approved):

prdlk (approved):

PR #728, opened 2022-11-01
Add support for more chains

ntindle:
I absolutely missed something here when setting this up


pkg/client/client.go
```diff
@@ -141,7 +161,14 @@ func (c *Client) GetIPFSApiAddress() string {
...
 	_, err := os.Stat(env_path)
 	if errors.Is(err, os.ErrNotExist) {
-		return IPFS_API_ADDRESS
+		switch c.clientMode {
+		case mt.ClientMode_ENDPOINT_BETA:
+			return IPFS_ADDRESS_BETA
+		case mt.ClientMode_ENDPOINT_DEV:
+			return IPFS_ADDRESS_DEV
+		case mt.ClientMode_ENDPOINT_LOCAL:
+			return IPFS_ADDRESS_LOCAL
```
prdlk, line 179: Add prod as an option for client mode here

pkg/client/client.go
```diff
@@ -124,7 +137,14 @@ func (c *Client) GetIPFSAddress() string {
 	// by default use .env if it exists
 	_, err := os.Stat(env_path)
 	if errors.Is(err, os.ErrNotExist) {
-		return IPFS_ADDRESS
+		switch c.clientMode {
+		case mt.ClientMode_ENDPOINT_BETA:
+			return IPFS_ADDRESS_BETA
+		case mt.ClientMode_ENDPOINT_DEV:
+			return IPFS_ADDRESS_DEV
+		case mt.ClientMode_ENDPOINT_LOCAL:
```
prdlk, line 152: Add prod as an option for client mode here

prdlk (changes requested):
The InitializeRequest from the Motor client supplies the mode that the node should use - pkg/motor uses this switch statement - so after adding that as a case this will work. Also build the protos using `make proto.go` so that the updated request contains the new prod option.

pkg/client/client.go
ntindle: done

pkg/client/client.go
ntindle: done

joshLong145 (approved):

prdlk (approved):

PR #729, opened 2022-11-01
Add Device Linking in Motor

mcjcloud:
Adds device linking in motor.

## Changes

- Added new `Connect` method for initializing motor's host
- Added new `OpenLinking` method to start listening for libp2p connections
- Added new `PairDevice` method for sending `AuthInfo` to a new device
- Unit test

<sub> Mention [stepsize] in a comment if you'd like to report some technical debt. See examples [here](https://app.stepsize.com/api/demo-pr-redirect).</sub>

PR #730, opened 2022-11-02
Add Device Onboarding to Motor

mcjcloud:
The companion PR to #729 

## Changes

- Add `OnboardDevice` method to motor
- Add custom callback with channels to TestSuite


bind/motor-mobile/connect.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	ct "github.com/sonr-io/sonr/third_party/types/common"
+	mt "github.com/sonr-io/sonr/third_party/types/motor/api/v1"
+	_ "golang.org/x/mobile/bind"
+)
+
+func Connect(buf []byte) ([]byte, error) {
```
joshLong145, line 11: Is there a reason the bullk of the implementation is in `bind` and not within `pkg/motor` for connection?

pkg/config/role.go
```diff
@@ -17,6 +17,9 @@ const (
 
 	// Role_HIGHWAY is for a Highway Node
 	Role_HIGHWAY
+
```
joshLong145: Perhaps we should call this roaming? linking is a bit unclear to me as to the state of the device.

pkg/motor/connect.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/sonr-io/sonr/pkg/host"
+	dp "github.com/sonr-io/sonr/pkg/motor/x/discover"
+	tp "github.com/sonr-io/sonr/pkg/motor/x/transmit"
+	ct "github.com/sonr-io/sonr/third_party/types/common"
+	mt "github.com/sonr-io/sonr/third_party/types/motor/api/v1"
+	v1 "github.com/sonr-io/sonr/third_party/types/motor/api/v1/service/v1"
+)
+
+func (mtr *motorNodeImpl) Connect(request mt.ConnectRequest) (*mt.ConnectResponse, error) {
+	if mtr.SonrHost != nil {
```
joshLong145: would there be a case when we would want to reconnect to a host after already establishing one?

pkg/motor/connect.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return &mt.PairingResponse{
+		Success: true,
+	}, nil
+}
+
+func (mtr *motorNodeImpl) connectToPeerAndTransmit(peerInfo peer.AddrInfo, request mt.PairingRequest) error {
+	if err := mtr.SonrHost.Connect(peerInfo); err != nil {
+		mtr.SonrHost.Host().Peerstore().ClearAddrs(peerInfo.ID)
```
joshLong145: does this reset the state of connection allowing for `connect` to be called again?

pkg/motor/connect.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to open stream while attempting to pair")
+	}
+	defer str.Close()
+
+	strWr := msgio.NewWriter(str)
+	authInfo := &ct.AuthInfo{
+		Address:   mtr.Address,
+		Did:       mtr.GetDID().String(),
+		AesPskKey: mtr.encryptionKey,
```
joshLong145: is this assuming the motor has already been logged in? we might want to check if the `encryptionKey` is nil

pkg/motor/connect.go
```diff
@@ -0,0 +1,206 @@
...
+		mtr.log.Warn("Host already connected")
+		return &mt.ConnectResponse{
+			Success: true,
+			Message: "Host already connected",
+		}, nil
+	}
+
+	// Setup host config
+	var err error
+	cnfg := config.DefaultConfig(config.Role_MOTOR, config.WithAccountAddress(mtr.GetAddress()), config.WithDeviceID(mtr.DeviceID), config.WithHomePath(mtr.homeDir), config.WithSupportPath(mtr.supportDir), config.WithTempPath(mtr.tempDir))
```
joshLong145: we should probably check that the motor has first been "logged in" before checking this.

pkg/motor/onboard.go
```diff
@@ -0,0 +1,68 @@
...
+		string(newShard[len(newShard)-3:]),
+		mtr.DeviceID,
+		string(cipherShard),
+	)
+	if err != nil {
+		return mt.OnboardDeviceResponse{}, fmt.Errorf("issue shard: %s", err)
+	}
+
+	// update the WhoIs with new vault
+	mtr.DIDDocument.AddService(service)
```
joshLong145, line 59: now that there are multiple valut services on documents do we need a way to contexualize them or is the `cid` enough?

pkg/config/role.go
mcjcloud: I decided to go ahead and get rid of this role, not sure why it was there.

pkg/motor/connect.go
mcjcloud: Good point. I'm adding a param to `ConnectRequest` which will allow you to skip this check explicitly.

pkg/motor/connect.go
mcjcloud: I think `Connect` can be called again either way, but this is just cleanup.

pkg/motor/connect.go
mcjcloud: Yes, you must be logged in in order to add a new device. Adding a check for this at the beginning of the function.

pkg/motor/connect.go
mcjcloud: Both devices need to do this, and the new device won't be logged in at this point.

pkg/motor/onboard.go
mcjcloud: Great catch. I added a method to remove the existing vault service.

joshLong145 (approved):

ntindle (approved):

PR #731, opened 2022-11-02
updates to schema proto coments for docs

joshLong145:
* Adds more comments for parsing into go docs in `schema` proto files


mcjcloud (approved):

prdlk (changes requested):
JS Style comments dont render in pkg.go.dev for our godoc, unsure if this will work can we test using protoc-doc-gen @ntindle ?

joshLong145:
updated comment structure, to use single comment structure.

prdlk (approved):

PR #762, opened 2022-11-15
Darshan/blok 105

lakshya-sky:
### Changes in Client API:
1. Use .env file for reading endpoints. This gives us better plug and play feature. 
2. added a method to query on chain account.

### Major changes on Wallet import
1. Use AccSeq from on-chain account
3. set wallet configs to have atleast threshold+1 config and use "psk" as a default participant on wallet import(earlier it was generating new shards)  

### Bug Fix:
1. bank shard with index 0 was not being stored on vault.


pkg/client/account.go
```diff
@@ -30,11 +35,44 @@ func (c *Client) CheckBalance(address string) (types.Coins, error) {
...
+	initClientCtx.Codec.UnmarshalJSON(accBz, &acc)
+	return &acc, nil
+}
+
 // RequestFaucet requests a faucet from the Sonr network
 func (c *Client) RequestFaucet(address string) error {
 	values := faucetRequest{
 		Address: address,
-		Coins:   []string{"12000snr"},
+		Coins:   []string{"120snr"},
```
mcjcloud: Can you verify that the motor unit tests still work with this change? I believe this was bumped up as a workaround for something...

pkg/crypto/mpc/wallet.go
```diff
@@ -26,12 +27,21 @@ type Wallet struct {
...
 	Threshold   int
+
+	AccSeq uint64
 }
 
 // GenerateWallet a new ECDSA private key shared among all the given participants.
 func GenerateWallet(cb common.MotorCallback, options ...WalletOption) (*Wallet, error) {
 	opt := defaultConfig()
 	w := opt.Apply(options...)
+	configLength := reflect.ValueOf(w.Configs).MapKeys()
```
mcjcloud: You shouldn't need to use reflection here,
```go
configLength := len(w.Configs)
```

pkg/did/internal/marshal/marshal.go
```diff
@@ -8,9 +8,9 @@ func NormalizeDocument(document []byte, normalizers ...Normalizer) ([]byte, erro
 	if err := json.Unmarshal(document, &tmp); err != nil {
 		return nil, err
 	}
-	for _, normalizer := range normalizers {
-		normalizer(tmp)
-	}
+	//for _, normalizer := range normalizers {
```
mcjcloud: Did you mean to leave this uncommented?

pkg/motor/login.go
```diff
@@ -122,33 +131,32 @@ func createWalletConfigs(id string, req mt.LoginWithKeysRequest, shards vault.Va
...
-		// build DSC Config if password is not provided
-		deviceShard, ok := shards.IssuedShards[id]
-		if !ok {
-			return nil, fmt.Errorf("could not find device shard with key '%s'", id)
-		}
-		dscShard, err := mpc.AesDecryptWithKey(req.AesDscKey, deviceShard)
-		if err != nil {
-			return nil, fmt.Errorf("error decrypting DSC shard: %s", err)
-		}
+	//if req.Password != "" {
```
mcjcloud: Did you mean to leave this commented out? The request is not guaranteed to have a password provided so this check is important.

pkg/motor/login.go
```diff
@@ -161,6 +169,33 @@ func createWalletConfigs(id string, req mt.LoginWithKeysRequest, shards vault.Va
 		return nil, fmt.Errorf("psk shard: %s", err)
 	}
 
+	bankShard0, err := mpc.AesDecryptWithKey(req.AesPskKey, shards.ShardBank[0])
```
mcjcloud: Decrypting all of these bank shards shouldn't be necessary. Providing just the two from above is enough to create the wallet.

pkg/motor/registry_test.go
```diff
@@ -3,68 +3,14 @@ package motor
...
-	"testing"
 	"time"
 
 	mtu "github.com/sonr-io/sonr/testutil/motor"
-	"github.com/sonr-io/sonr/third_party/types/common"
 	mt "github.com/sonr-io/sonr/third_party/types/motor/api/v1"
 	"github.com/stretchr/testify/assert"
 )
 
-func (suite *MotorTestSuite) Test_LoginWithKeys() {
```
mcjcloud: Why was this removed?

pkg/motor/schema_test.go
```diff
@@ -99,6 +99,7 @@ func (suite *MotorTestSuite) Test_QuerySchemaByCreator() {
 	}
 }
 
+/*
```
mcjcloud: Why is this commented out?

pkg/motor/create_account.go
```diff
@@ -8,14 +8,13 @@ import (
...
 	"github.com/sonr-io/sonr/pkg/did/ssi"
-	"github.com/sonr-io/sonr/pkg/tx"
 	"github.com/sonr-io/sonr/pkg/vault"
 	"github.com/sonr-io/sonr/third_party/types/common"
 	mt "github.com/sonr-io/sonr/third_party/types/motor/api/v1"
 	rt "github.com/sonr-io/sonr/x/registry/types"
 )
 
-func (mtr *motorNodeImpl) CreateAccount(request mt.CreateAccountRequest) (mt.CreateAccountResponse, error) {
+func (mtr *motorNodeImpl) CreateAccount(request mt.CreateAccountRequest, waitForVault bool) (mt.CreateAccountResponse, error) {
```
mcjcloud: Take a look at `pkg/motor/connect_test.go` for how unit tests can be done without adding this test boolean.

mcjcloud (changes requested):

pkg/motor/create_account.go
```diff
@@ -117,22 +116,26 @@ func (mtr *motorNodeImpl) CreateAccountWithKeys(request mt.CreateAccountWithKeys
 	mtr.triggerWalletEvent(common.WalletEvent{Type: common.WALLET_EVENT_TYPE_DID_DOCUMENT_CREATE_ERROR})
 	mtr.triggerWalletEvent(common.WalletEvent{Type: common.WALLET_EVENT_TYPE_SHARD_GENERATE_START})
 
-	go createVault(mtr, request)
+	r := make(chan int)
```
joshLong145: nit: we shouldn't make the channel unless its flagged to wait

